PR #6, opened 2021-07-07
Add admin:DownloadRaw action

klauspost:
Required for upcoming feature.

This PR will be linked.

### DIFF
--- a/iam/policy/admin-action.go
+++ b/iam/policy/admin-action.go
@@ -56,8 +56,8 @@ const (
 	HealthInfoAdminAction = "admin:OBDInfo"
 	// BandwidthMonitorAction - allow monitoring bandwidth usage
 	BandwidthMonitorAction = "admin:BandwidthMonitor"
-	// DownloadRawAction - allows downloading raw files from backend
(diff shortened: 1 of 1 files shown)


iam/policy/admin-action.go
```diff
@@ -56,6 +56,8 @@ const (
 	HealthInfoAdminAction = "admin:OBDInfo"
 	// BandwidthMonitorAction - allow monitoring bandwidth usage
 	BandwidthMonitorAction = "admin:BandwidthMonitor"
+	// DownloadRawAction - allows downloading raw files from backend
+	DownloadRawAction = "admin:DownloadRaw"
```
harshavardhana, line 60: shouldn't this be something like InspectObject @klauspost ?

harshavardhana (changes requested):

iam/policy/admin-action.go
klauspost: @harshavardhana Dunno, it will not inspect anything, just download the file you ask for from all servers.

harshavardhana (approved):

PR #9, opened 2021-09-27
New admin actions for cluster replication feature

donatello:


vadmeste (approved):
LGTM

iam/policy/admin-action.go
```diff
@@ -85,6 +85,15 @@ const (
 	// GetUserAdminAction - allows GET permission on user info
 	GetUserAdminAction = "admin:GetUser"
 
+	// Cluster Replicate Actions
+
+	// ClusterReplicateAddAction - allow adding replication clusters
+	ClusterReplicateAddAction = "admin:ClusterReplicateAdd"
```
vadmeste, line 91: It looks like 'Manage' is better than 'Add' here since this action is already used in multiple cases

vadmeste (changes requested):

iam/policy/admin-action.go
harshavardhana: Yeah makes sense @vadmeste

donatello: We use a single set of service credentials to perform all replication management actions on peers such as creating/removing bucket, making service account updates etc. This service account would likely need only one special permission to perform all these actions - I wasn't sure of a name for this, but I think we can use `ClusterReplicateManage`. Will send an update to the PRs.

donatello:
I don't expect we will need further permissions - so we can take this in.

iam/policy/admin-action.go
harshavardhana: Do we still need `Add` if we have Manage? @donatello ?

donatello: Yes, add is to add the cluster repl config, manage is to perform internal replication operations via the svc account created for this purpose. Manage permission just isolates actions by this service account.

harshavardhana (approved):

iam/policy/admin-action.go
```diff
@@ -85,6 +85,18 @@ const (
...
+
+	// SiteReplicationAddAction - allow adding clusters for site-level replication
+	SiteReplicationAddAction = "admin:SiteReplicationAdd"
+	// SiteReplicationDisableAction - allow disabling a cluster from replication
+	SiteReplicationDisableAction = "admin:SiteReplicationDisable"
+	// SiteReplicationInfoAction - allow getting site replication info
+	SiteReplicationInfoAction = "admin:SiteReplicationInfo"
+	// SiteReplicationOperationAction - allow performing site replication
+	// create/update/delete operations to peers
+	SiteReplicationOperationAction = "admin:SiteReplicationManage"
```
vadmeste, line 98: s/SiteReplicationManage/SiteReplicationOperation/

vadmeste (approved):

vadmeste (changes requested):
One comment

iam/policy/admin-action.go
harshavardhana, line 98:
```suggestion
	SiteReplicationOperationAction = "admin:SiteReplicationOperation"
```

harshavardhana (changes requested):

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #12, opened 2021-10-25
Add function to verify cluster specific license

anjalshireesh:
Apart from vefifying that the license token is valid, this function will
also validate that it belongs to the given cluster (deployment id)


anjalshireesh:
Closing the PR as we're now planning to start using an `api key` instead of `license` for communication with subnet.

PR #15, opened 2021-11-03
Fix name of admin policy to `consoleAdmin`

donatello:


harshavardhana (approved):

vadmeste (commented):
I wonder if there is any issue associated to his @donatello ?

donatello:
> I wonder if there is any issue associated to his @donatello ?

No, the change that introduced it, is not yet in master. It is for https://github.com/minio/minio/pull/13567

PR #17, opened 2021-11-16
Add certificate reload functions

klauspost:
Allows reload on OS signal or manual reloads.

Some simplifications to existing code added.

First part of https://github.com/minio/minio/issues/13662

PR #18, opened 2021-11-16
certs: reduce reload time for symlinks to 15 min

aead:
This commit reduces the certificate reload time
window (for symlinks) to 15 min. Before, certificates
would have been reloaded every 24h - which can be
an issue when certificates are valid for less than 24h.

Now, certificates mounted as symlinks will be reloaded
after 15 minutes - which is on the one side frequent enough
to avoid expiry issues and on the other side rare enough
to not cause any noticeable I/O load.

### DIFF
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -187,7 +187,7 @@ func (m *Manager) watchSymlinks(certFile, keyFile string) {
 		select {
 		case <-m.ctx.Done():
 			return // Once stopped exits this routine.
-		case <-time.After(24 * time.Hour):
+		case <-time.After(15 * time.Minute):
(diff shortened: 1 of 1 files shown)


harshavardhana:
Which PR is correct looks like there are two #17

klauspost:
I reduced it to 1 hour in #17 - but will change it to 15m.

aead:
#17 Adds a signal-based notify mechanism. So, you can reload certificates on arbitrary signals.
However, there may be no signal for an event that changes the symlink destination. E.g. updating the K8S secret.
So, both solutions have benefits. 

#17 Reloads certificates instantly **if there is a signal we are listing on**
#18 Reloads certificates every 15 minutes

aead:
In favour of #17

PR #22, opened 2022-01-04
Consider SID in merging and equality of policy statements

bluemmb:
This does 2 things:
- Adds SID to cloned statements in **bucket** and **iam** policy merge methods
- Considers SID in checking equality of policy statements

Should solve [minio#13905](https://github.com/minio/minio/issues/13905)

### DIFF
--- a/iam/policy/policy.go
+++ b/iam/policy/policy.go
@@ -223,10 +223,14 @@ func (iamp Policy) Merge(input Policy) Policy {
 		mergedPolicy.Version = input.Version
 	}
 	for _, st := range iamp.Statements {
-		mergedPolicy.Statements = append(mergedPolicy.Statements, st.Clone())
+		var newSt Statement = st.Clone()
+		newSt.SID = st.SID
+		mergedPolicy.Statements = append(mergedPolicy.Statements, newSt)
 	}
 	for _, st := range input.Statements {
-		mergedPolicy.Statements = append(mergedPolicy.Statements, st.Clone())
+		var newSt Statement = st.Clone()
+		newSt.SID = st.SID
+		mergedPolicy.Statements = append(mergedPolicy.Statements, newSt)
 	}
 	mergedPolicy.dropDuplicateStatements()
 	return mergedPolicy
--- a/bucket/policy/policy.go
+++ b/bucket/policy/policy.go
@@ -112,10 +112,14 @@ func (policy Policy) Merge(input Policy) Policy {
 		mergedPolicy.Version = input.Version
 	}
 	for _, st := range policy.Statements {
-		mergedPolicy.Statements = append(mergedPolicy.Statements, st.Clone())
+		var newSt Statement = st.Clone()
+		newSt.SID = st.SID
+		mergedPolicy.Statements = append(mergedPolicy.Statements, newSt)
 	}
 	for _, st := range input.Statements {
-		mergedPolicy.Statements = append(mergedPolicy.Statements, st.Clone())
+		var newSt Statement = st.Clone()
+		newSt.SID = st.SID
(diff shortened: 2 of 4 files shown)


iam/policy/policy.go
```diff
@@ -223,10 +223,14 @@ func (iamp Policy) Merge(input Policy) Policy {
...
-		mergedPolicy.Statements = append(mergedPolicy.Statements, st.Clone())
+		var newSt Statement = st.Clone()
+		newSt.SID = st.SID
+		mergedPolicy.Statements = append(mergedPolicy.Statements, newSt)
 	}
 	for _, st := range input.Statements {
-		mergedPolicy.Statements = append(mergedPolicy.Statements, st.Clone())
+		var newSt Statement = st.Clone()
+		newSt.SID = st.SID
+		mergedPolicy.Statements = append(mergedPolicy.Statements, newSt)
```
harshavardhana, line 233: It is really debatable which Sid is valid here - my assumption here is that we should preserve the Sid that is `inline` instead of blindly replacing it with any Sid that comes along.

This type of replacement has a bug where it will still become empty if you have other Statements without Sid.

harshavardhana (changes requested):

iam/policy/policy.go
bluemmb: @harshavardhana Thanks for checking my PR

I'm not sure that by `inline` you mean the policy that user gives us while creating a new Service Account?

If that's the case `input` comes later in `mergedPolicy` and overriders previous ones in dropping duplicate statements process.

About the second concern, I think you mean if two statements are same but the second one has no Sid, it will override the one that has. On that matter, maybe we could always pass the Sid forward when it comes to that situation, something like this:

```go
func (iamp *Policy) dropDuplicateStatements() {
redo:
	for i := range iamp.Statements {
		for j, statement := range iamp.Statements[i+1:] {
			if !iamp.Statements[i].Equals(statement) {
				continue
			}
			if iamp.Statements[i+1+j].SID == "" && iamp.Statements[i].SID != "" {
				iamp.Statements[i+1+j].SID = iamp.Statements[i].SID
			}
			iamp.Statements = append(iamp.Statements[:i], iamp.Statements[i+1:]...)
			goto redo
		}
	}
}
```

Pardon me if my understanding is wrong in any of these, since I have just recently started using minio and have very small knowledge on it's source code 🙏

harshavardhana: The bug is actually not here in preserving the `Sid` instead the way service account was returning an incorrect Policy by doing an unnecessary Merge.

harshavardhana:
also two Sid's with same name doesn't mean they are equal - Sid is just a comment and comments can match and we shouldn't incorrectly match them as same without evaluating the Statements and actions. 

I think this PR is not needed the fix is on the server side.

bluemmb:
> also two Sid's with same name doesn't mean they are equal - Sid is just a comment and comments can match and we shouldn't incorrectly match them as same without evaluating the Statements and actions.
> 
> I think this PR is not needed the fix is on the server side.

@harshavardhana Thanks for checking this and the PR you have created on minio repo, solving this issue will help us a lot.

I just wanted to ask that in [s3 docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_sid.html) it mentions that Sid is an identifier and should be unique. It had got me into thinking that it might be more than a comment?

harshavardhana:
> I just wanted to ask that in s3 docs it mentions that Sid is an identifier and should be unique. It had got me into thinking that it might be more than a comment?

Sid is unique for a statement but it's optional it has no realistic meaning outside of this.

It won't be optional if it's meaningful in some way. It is basically just a comment at best.

PR #24, opened 2022-01-14
Add admin action for removing site replication

poornas:


iam/policy/admin-action.go
```diff
@@ -91,8 +91,8 @@ const (
 
 	// SiteReplicationAddAction - allow adding clusters for site-level replication
 	SiteReplicationAddAction = "admin:SiteReplicationAdd"
-	// SiteReplicationDisableAction - allow disabling a cluster from replication
-	SiteReplicationDisableAction = "admin:SiteReplicationDisable"
+	// SiteReplicationRemoveAction - allow removing a cluster from replication
+	SiteReplicationRemoveAction = "admin:SiteReplicationRemove"
```
harshavardhana, line 95: We can't remove actions in this manner @poornas you should leave the Disable as is only add new actions as needed. 

Also supportedAdminActions is missing most of the Site* calls they need to be added.

harshavardhana (changes requested):

harshavardhana (approved):

harshavardhana:
PTAL @donatello

donatello (approved):

PR #25, opened 2022-01-18
Increase lookupEnv timeout to 6.5 seconds

dvaldivia:
On certain environments, 3 second timeout is too small causing MinIO to panic and fail complaining the environment could't be resolved

Signed-off-by: Daniel Valdivia <18384552+dvaldivia@users.noreply.github.com>

### DIFF
--- a/env/web_env.go
+++ b/env/web_env.go
@@ -90,7 +90,7 @@ func getEnvValueFromHTTP(urlStr, envKey string) (string, string, string, error) 
 		return "", "", "", err
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 6.5*time.Second)
 	defer cancel()
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, envURL+"?key="+envKey, nil)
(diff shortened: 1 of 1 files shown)


env/web_env.go
```diff
@@ -90,7 +90,9 @@ func getEnvValueFromHTTP(urlStr, envKey string) (string, string, string, error)
 		return "", "", "", err
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	// Adding a timeout of 6.5 seconds to deal with k3s slow dns resolution caused in turn by
+	// CoreDNS 6 second default timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), 6.5*time.Second)
```
kannappanr, line 95: should this be `6*time.Second`?

harshavardhana (approved):

harshavardhana (approved):

env/web_env.go
vadmeste: If this is only called once - this is what it looks like - you can even set it to one minute.

harshavardhana: We don't want our CLI to blocked for such a long time @vadmeste this is specifically to fix a k3s DNS resolution bug.

PR #26, opened 2022-02-10
Add madmin:Bucket condition for admin policy

vadmeste:
This will be used with admin:{Set,Get}BucketTarget action to limit an
account to be able to access to the remote bucket configuration of
one or more buckets.


harshavardhana:
@vadmeste is there a plan to complete this? any ETA?

harshavardhana:
Closing this we can bring this back when needed.

PR #34, opened 2022-07-22
allow valid tagging conditionals to various API actions

harshavardhana:
refer https://github.com/minio/minio/issues/10265

This change is needed before we get server side support


bucket/policy/action.go
```diff
@@ -336,7 +336,6 @@ func createActionConditionKeyMap() map[Action]condition.KeySet {
 			append([]condition.Key{
 				condition.S3XAmzServerSideEncryption.ToKey(),
 				condition.S3XAmzServerSideEncryptionCustomerAlgorithm.ToKey(),
-				condition.ExistingObjectTag.ToKey(),
```
harshavardhana, line 339: removing it here since commonKeys already carries `ExistingObjectTag`

PR #38, opened 2022-08-31
Add IAM policies  to control access to minio handlers that will communicate with KES

reivaj05:
Adds IAM policies  to control access to minio handlers that will communicate with KES.

The handlers will be added in minio in another PR in which i'm currently working on.

Please let me know if there are more places where i need to make changes since this is the first time i work with policies.

Closes: https://github.com/miniohq/engineering/issues/924

### DIFF
--- a/iam/policy/admin-action.go
+++ b/iam/policy/admin-action.go
@@ -53,6 +53,32 @@ const (
 	ConsoleLogAdminAction = "admin:ConsoleLog"
 	// KMSCreateKeyAdminAction - allow creating a new KMS master key
 	KMSCreateKeyAdminAction = "admin:KMSCreateKey"
+	// KMSDeleteKeyAdminAction - allow deleting a KMS master key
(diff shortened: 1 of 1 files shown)


cniackz (approved):
I don't know of any other places, but this change looks good to me.
Let's wait for Daniel, Lenin and Senior guys to see if they know more of it to help here.

Alevsk (approved):

Alevsk:
Does this looks good @harshavardhana  ?

harshavardhana:
are we going to implement these handlers at MinIO or Kes @Alevsk @reivaj05 ?

harshavardhana (changes requested):
Let's discuss this first - if we are going to do this then I would suggest we add `kms:CreateKey` like actions instead of `admin:*` 

This allows us to have special Conditions and have all the S3 API-related functionality.

reivaj05:
no need

PR #39, opened 2022-09-13
Add IAM policies  to control access to KES APIs

reivaj05:
Adds IAM policies to control access to minio handlers that will communicate with KES.

The handlers will be added in minio in another PR in which i'm currently working on.

Please let me know if there are more places where i need to make changes since this is the first time i work with policies.

Closes: https://github.com/miniohq/engineering/issues/924

### DIFF
--- a/iam/policy/kms-action.go
+++ b/iam/policy/kms-action.go
@@ -1,0 +1,84 @@
+// Copyright (c) 2015-2022 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


iam/policy/kms-action.go
```diff
@@ -0,0 +1,84 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package iampolicy
+
+// KMSAction - KMS policy action.
+type KMSAction string
+
+const (
+	// KMSCreateKeyAction - allow creating a new KMS master key
+	KMSCreateKeyAction = "kms:KMSCreateKey"
```
harshavardhana, line 25: we need to just keep 
```suggestion
	KMSCreateKeyAction = "kms:CreateKey"
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #42, opened 2022-09-28
Update console admin policy and allowed kms:* actions to be processed

reivaj05:
Update console admin policy and allowed kms:* actions to be processed


reivaj05:
@aead PTAL

PR #49, opened 2022-11-30
Bump up madmin-go to v2

donatello:
Not sure if the go.sum change needs to be as large as it is. Ran `go mod tidy` already.


harshavardhana (changes requested):
You need to run `go mod tidy -compat=1.18` @donatello

harshavardhana (approved):

PR #52, opened 2022-12-10
Add LDAP SRV Record lookup support

donatello:
Adds `SRVRecordName` configuration parameter and validation for it. The
value can be `ldaps`, `ldap` or the special value `on`. This value is
combined with the `ServerAddr` parameter to lookup a DNS SRV record for
finding target LDAP servers.

The SRV record name format is always `_service._proto.domainname`. For LDAP
`proto` value here is always `tcp`. The `service` value is usually
`ldap`, and sometimes `ldaps`. When `SRVRecordName` is set to `ldap` or
`ldaps`, this value is substituted for `service`. The `domainname` value
is given by the `ServerAddr` parameter.

When a different SRV record name needs to be given, set `SRVRecordName`
to the special value `on`, and set the `ServerAddr` to the full SRV
record name, e.g. `_ldapish._tcpish.myldapserver.com`.

### DIFF
--- a/ldap/ldap.go
+++ b/ldap/ldap.go
@@ -41,6 +41,7 @@ type Config struct {
 
 	// E.g. "ldap.minio.io:636"
 	ServerAddr     string
+	SRVRecordName  string
 	TLSSkipVerify  bool // allows skipping TLS verification
 	ServerInsecure bool // allows plain text connection to LDAP server
 	ServerStartTLS bool // allows using StartTLS connection to LDAP server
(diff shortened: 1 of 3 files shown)


donatello:
This can be taken after https://github.com/minio/pkg/pull/51

donatello:
DNS SRV ref for LDAP https://ldap.com/dns-srv-records-for-ldap/

ldap/ldap.go
```diff
@@ -104,6 +94,71 @@ func (l *Config) Connect() (ldapConn *ldap.Conn, err error) {
...
+			}
+		}
+
+		return l.connect(ldapAddr)
+	}
+
+	// SRV Record lookup is enabled.
+	_, addrs, err := net.LookupSRV(srvService, srvProto, srvName)
+	if err != nil {
+		return nil, fmt.Errorf("DNS SRV Record lookup error: %v", err)
```
vadmeste, line 137: s/%v/%w

ldap/ldap.go
```diff
@@ -104,6 +94,71 @@ func (l *Config) Connect() (ldapConn *ldap.Conn, err error) {
...
+			} else {
+				return nil, err
+			}
+		}
+
+		return l.connect(ldapAddr)
+	}
+
+	// SRV Record lookup is enabled.
+	_, addrs, err := net.LookupSRV(srvService, srvProto, srvName)
```
vadmeste, line 135: Just wanted a confirmation, does net.LookupSRV() work if srvServce == "" and srvProto == "" ? this is when SRVRecordName == "on"

vadmeste (commented):
Comments

ldap/ldap.go
donatello: Yes, this is the documented behavior - https://pkg.go.dev/net#LookupSRV

vadmeste (approved):
LGTM

krisis (approved):

PR #53, opened 2022-12-11
NotAction Policy element added

mahi072:
Changes made in Action file to make the "NotAction" reflect on "Allow".
Please review this.


harshavardhana:
@mahi072 this is not the correct fix.

mahi072:
> @mahi072 this is not the correct fix.

Can you please tell me where to make actual changes?
Are the changes which I made is in right source file?, so that I can proceed with the further corrections.

harshavardhana:
> Can you please tell me where to make actual changes?

@mahi072 you need to add a new type of `Action`  called `NotAction` that serves as the inverse of `Action` - look at `action.go` file.

PR #54, opened 2022-12-29
feat(iam/policy): support NotAction field

eirture:
I'm trying to fix minio/minio#15328 to implement `NotAction` field in a policy statement.

### DIFF
--- a/iam/policy/statement.go
+++ b/iam/policy/statement.go
@@ -29,6 +29,7 @@ type Statement struct {
 	SID        policy.ID           `json:"Sid,omitempty"`
 	Effect     policy.Effect       `json:"Effect"`
 	Actions    ActionSet           `json:"Action"`
+	NotActions ActionSet           `json:"NotAction,omitempty"`
 	Resources  ResourceSet         `json:"Resource,omitempty"`
 	Conditions condition.Functions `json:"Condition,omitempty"`
 }
@@ -36,7 +37,8 @@ type Statement struct {
 // IsAllowed - checks given policy args is allowed to continue the Rest API.
 func (statement Statement) IsAllowed(args Args) bool {
 	check := func() bool {
-		if !statement.Actions.Match(args.Action) {
+		if (!statement.Actions.Match(args.Action) && !statement.Actions.IsEmpty()) ||
+			statement.NotActions.Match(args.Action) {
 			return false
 		}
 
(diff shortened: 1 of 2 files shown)


eirture:
Hi @harshavardhana , I'm trying to implement NotAction in a policy statement. PTAL thx! 🙏

krisis (approved):

iam/policy/statement.go
```diff
@@ -166,8 +171,14 @@ func (statement Statement) Equals(st Statement) bool {
...
-	return NewStatement(statement.SID, statement.Effect, statement.Actions.Clone(),
-		statement.Resources.Clone(), statement.Conditions.Clone())
+	return Statement{
+		SID:        statement.SID,
+		Effect:     statement.Effect,
+		Actions:    statement.Actions,
+		NotActions: statement.NotActions,
+		Resources:  statement.Resources,
+		Conditions: statement.Conditions,
+	}
```
harshavardhana, line 181: why is this being changed? @eirture

harshavardhana (changes requested):

harshavardhana (changes requested):
We need more tests here to cater for follow styles

- "Allow" with NotAction 
- "Deny" with NotAction

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notaction.html

iam/policy/statement.go
eirture: This method is a clone operation. A new field `NoActions` add to the statement, when we clone a statement we should copy the `NotActions` too. Is this right?

eirture:
> We need more tests here to cater for follow styles
> 
> * "Allow" with NotAction
> * "Deny" with NotAction
> 
> https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notaction.html

@harshavardhana  thx for review. I will add more tests for this case.

iam/policy/statement.go
harshavardhana: yes but you need to enhance the NewStatement() to take `NotActions.Clone()` as input what is happening right now is not a copy AFAIK.

eirture: yes but I think the NewStatement function is a promise to users, we can't change parameters anymore. Can we add a new function to create a statement? Could you give me some advice?

harshavardhana: yes you can of course add, the only consumer of this function is MinIO - we don't expect users to use this package outside of MinIO. That is out scope for us.

harshavardhana: you can also add NewStatementWithNotActions() if you want to keep previous function as is.

eirture:
Hi @harshavardhana @krisis, I have pushed some updates about review comments. PTAL again, thx! 🍺

harshavardhana (approved):

PR #59, opened 2023-02-27
Add units in the audit log header

nitisht:
This allows removing the `ns` from the field, allowing better plots of these fields in log management systems.


logger/message/audit/entry.go
```diff
@@ -35,17 +35,19 @@ type Entry struct {
...
+		Bucket                  string          `json:"bucket,omitempty"`
+		Object                  string          `json:"object,omitempty"`
+		Objects                 []ObjectVersion `json:"objects,omitempty"`
+		Status                  string          `json:"status,omitempty"`
+		StatusCode              int             `json:"statusCode,omitempty"`
+		InputBytes              int64           `json:"rx"`
+		OutputBytes             int64           `json:"tx"`
+		HeaderBytes             int64           `json:"txHeaders,omitempty"`
+		TimeToFirstByte         string          `json:"timeToFirstByte,omitempty"`
+		TimeToFirstByteWithUnit string          `json:"timeToFirstByte(ns),omitempty"`
```
Praveenrajmani, line 48:
```suggestion
		TimeToFirstByteInNS string          `json:"timeToFirstByteInNS,omitempty"`
```

logger/message/audit/entry.go
```diff
@@ -35,17 +35,19 @@ type Entry struct {
...
+		TimeToResponse          string          `json:"timeToResponse,omitempty"`
+		TimeToResponseWithUnit  string          `json:"timeToResponse(ns),omitempty"`
```
Praveenrajmani, line 50:
```suggestion
		TimeToResponseInNS  string          `json:"timeToResponseInNS,omitempty"`
```

Praveenrajmani (commented):
can we indicate that the unit here is `ns` in the field names?

nitisht:
> can we indicate that the unit here is `ns` in the field names?

Yes, this is actually better. Fixing now. Thanks @Praveenrajmani

logger/message/audit/entry.go
```diff
@@ -35,17 +35,19 @@ type Entry struct {
...
+		Objects                 []ObjectVersion `json:"objects,omitempty"`
+		Status                  string          `json:"status,omitempty"`
+		StatusCode              int             `json:"statusCode,omitempty"`
+		InputBytes              int64           `json:"rx"`
+		OutputBytes             int64           `json:"tx"`
+		HeaderBytes             int64           `json:"txHeaders,omitempty"`
+		TimeToFirstByte         string          `json:"timeToFirstByte,omitempty"`
+		TimeToFirstByteInNS string          `json:"timeToFirstByteInNS,omitempty"`
+		TimeToResponse          string          `json:"timeToResponse,omitempty"`
+		TimeToResponseInNS  string          `json:"timeToResponseInNS,omitempty"`
```
Praveenrajmani, line 50: we need to run `go fmt` on this file

Praveenrajmani (approved):
Just a minor nit. LGTM otherwise

logger/message/audit/entry.go
nitisht: done

nitisht:
cc @harshavardhana could you please take a look..

PR #62, opened 2023-03-28
Add utility functions to check for specific connection errors

Praveenrajmani:
These utility functions are required for a change in minio/minio (will add the PR reference soon)

### DIFF
--- a/net/url.go
+++ b/net/url.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"syscall"
 )
 
 // URL - improved JSON friendly url.URL.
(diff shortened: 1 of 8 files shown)


Praveenrajmani:
looks like the lint pkgs have been upgraded 

```
golangci/golangci-lint info found version: 1.52.2 for v1.52.2/linux/amd64
```

will fix this in a separate PR

net/url.go
```diff
@@ -251,3 +252,21 @@ func IsNetworkOrHostDown(err error, expectTimeouts bool) bool {
...
+	return errors.Is(err, syscall.ECONNRESET)
+}
+
+// IsConnRefusedErr - Checks for "connection refused" errors.
+func IsConnRefusedErr(err error) bool {
+	if strings.Contains(err.Error(), "connection refused") {
+		return true
+	}
+	// incase if error message is wrapped.
+	return errors.Is(err, syscall.ECONNREFUSED)
```
balamurugana, line 271: When wouldn't the error type of syscall? You would need to deal with `errors.Unwrap()` if needed

Praveenrajmani: Just because `errors.Is` traverses and catches the `syscall.Error`, it is used. `errors.Unwrap()` does only one level check right and it won't go beyond that?

Also, this function is moved from `minio/minio` for common usage.

Praveenrajmani: > When wouldn't the error type of syscall

These utils were being used in event store mechanism

harshavardhana (approved):

balamurugana (approved):

PR #64, opened 2023-04-27
Add request path and Host and type to the audit

vadmeste:
- Request path and Host are interesting information for troubleshooting
- Type to specify what type of audit messages is this, it can be S3 operations, admin management, bucket management. It would be easier with this to filter audit logs

### DIFF
--- a/logger/message/audit/entry.go
+++ b/logger/message/audit/entry.go
@@ -31,6 +31,10 @@ type Entry struct {
 	DeploymentID string    `json:"deploymentid,omitempty"`
 	Time         time.Time `json:"time"`
 	Event        string    `json:"event"`
+
+	// Class of audit message - S3, admin ops, bucket management
+	Type string `json:"type,omitempty"`
+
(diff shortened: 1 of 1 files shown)


logger/message/audit/entry.go
```diff
@@ -31,6 +31,10 @@ type Entry struct {
 	DeploymentID string    `json:"deploymentid,omitempty"`
 	Time         time.Time `json:"time"`
 	Event        string    `json:"event"`
+
+	// Class of audit message - S3, admin ops, bucket management
+	Category string `json:"category,omitempty"`
```
harshavardhana, line 36: Should this be a typed category? We can perhaps call this Type? Like AuditType ?

harshavardhana (changes requested):

logger/message/audit/entry.go
vadmeste: > Should this be a typed category? We can perhaps call this Type? Like AuditType ?

Done

harshavardhana (approved):

PR #71, opened 2023-06-22
Speed up duplicate policy statement removal

donatello:
Duplicate detection now always runs in `O(n^2)` time (n = number of statements). Previously this could take `O(n^3)` time (when the number of duplicates is high).


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/fatih/color v1.14.1
 	github.com/fatih/structs v1.1.0
 	github.com/go-ldap/ldap/v3 v3.4.4
-	github.com/lestrrat-go/jwx v1.2.25
+	github.com/lestrrat-go/jwx v1.2.26
```
harshavardhana, line 10: @donatello send a separate PR for dependency update?

donatello: ok sure

donatello: This is done now, but CI will fail without the security fix.

harshavardhana (approved):

poornas (approved):

PR #74, opened 2023-06-26
Catch up action set implementation in policy pkgs

donatello:
Goal here is to identify the differences in the implementations of the two iam/policy and bucket/policy packages and eventually merge them into a single package without too much code duplication


donatello:
With this change, action set implementation will be same in the two policy packages, with the exception that iam/policy package has some extra methods to deal with admin and kms actions.

harshavardhana (approved):

PR #75, opened 2023-06-26
fix: bucket matching for bucket policy resource

donatello:
This fixes some corner cases in bucket matching in bucket policy
resources.

Failing tests for bucket/policy resource in existing implementation are
added.

This happens because we cannot parse out a unique bucket pattern from a
resource string in all cases: when resource pattern is
`mybucket*a/myobject*`, it should match `mybucket/a/myobject`.

For this pattern, the buckets matched are:
   1. `mybucket` (e.g. resource value is "mybucket/a/myobject"),
   2. `mybucketXa`, where X is a string not containing a `/` (e.g.
   resource value is "mybucket-east-a/myobject"), and
   3. `mybucketY`, where Y is a string not containing a `/` (e.g.
   resource value that matches is "mybucket22/a/myobject")

To fix this a new MatchAsPatternPrefix function added to wildcard
package and bucket name matching code is updated.

### DIFF
--- a/bucket/policy/resource.go
+++ b/bucket/policy/resource.go
@@ -90,13 +90,23 @@ func (r *Resource) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// Validate - validates Resource is for given bucket or not.
+// Validate - validates that given bucketName is matched by Resource.
 func (r Resource) Validate(bucketName string) error {
 	if !r.IsValid() {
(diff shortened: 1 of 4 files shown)


donatello:
Developed on top of https://github.com/minio/pkg/pull/74 so that needs to be merged first.

harshavardhana (approved):

poornas (approved):

PR #78, opened 2023-06-27
Fix: statement validation in iam/policy pkg

donatello:


iam/policy/statement.go
```diff
@@ -127,8 +127,14 @@ func (statement Statement) isValid() error {
...
-			return Errorf("unsupported Resource found %v for action %v", statement.Resources, action)
+		if action.isObjectAction() {
+			if !statement.Resources.objectResourceExists() {
+				return Errorf("unsupported Resource found %v for action %v", statement.Resources, action)
+			}
+		} else {
+			if !statement.Resources.bucketResourceExists() {
+				return Errorf("unsupported Resource found %v for action %v", statement.Resources, action)
+			}
 		}
```
donatello, line 138: This change is from the same validation implementation for `bucket/policy/statement.go`

donatello:
This is pending a discussion with @abperiasamy but it is not stopping other PRs from going through.

harshavardhana (changes requested):

bucket/policy/statement.go
```diff
@@ -98,10 +98,18 @@ func (statement Statement) isValid() error {
 		return Errorf("Action must not be empty")
 	}
 
+	if !statement.SID.IsValid() {
```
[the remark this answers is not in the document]
donatello: No, it passes for empty strings:

```
// IsValid - checks if ID is valid or not.
func (id ID) IsValid() bool {
	return utf8.ValidString(string(id))
}
```
https://go.dev/play/p/vFeEIAN1MuW

iam/policy/statement.go
```diff
@@ -127,8 +127,14 @@ func (statement Statement) isValid() error {
...
 
 	for action := range statement.Actions {
-		if !statement.Resources.objectResourceExists() && !statement.Resources.bucketResourceExists() {
-			return Errorf("unsupported Resource found %v for action %v", statement.Resources, action)
+		if action.isObjectAction() {
+			if !statement.Resources.objectResourceExists() {
+				return Errorf("unsupported Resource found %v for action %v", statement.Resources, action)
+			}
+		} else {
+			if !statement.Resources.bucketResourceExists() {
```
vadmeste, line 135: I am not sure - we have some admin:* actions as well, and I don't think they have bucket resources.. wouldn't that fail in that case @donatello ? (basically adding a IAM policy which contains admin:* actions

vadmeste (commented):
Question

iam/policy/statement.go
donatello: This is statement level validation - for admin requests it is validated earlier in line 95, using a different logic.

donatello:
After discussing with @abperiasamy the plan here is to generate warnings instead of rejecting the policy where the resource values are inappropriate for bucket/iam policy statements, so will be closing this one.

PR #79, opened 2023-06-28
Move common policy code into iam/policy

donatello:
This change basically removes duplicate code used in bucket/policy
implementation:

- Move ID and Effect implementation to iam/policy package

- Move Action and ActionSet implementation to iam/policy package

- Move Resource and ResourceSet implementation to iam/policy
package (added a separate `.ValidateBucket(bucketName)` for use with
bucket/policy).

- Move condition subpackage to iam/policy

Also JSON serialization for bucket policy will not perform any
validation of the policy itself similar to the iam policy
implementation. Validation should be after deserialization by caller.


donatello:
Built on top of https://github.com/minio/pkg/pull/77 so we need to merge that one first.

donatello:
PTAL this is ready

harshavardhana (approved):

donatello:
This is a useful code cleanup refactor, this can be taken in imo. PTAL @krisis

krisis (approved):

poornas (approved):

PR #81, opened 2023-06-29
Merge policy packages

donatello:
- Move Principal implementation into iam/policy

- Rename bucket/policy Policy and Statement types to BucketPolicy and
BPStatement respectively. Some functions are also renamed to avoid name
clashes.

- Renamed types mentioned above are moved into iam/policy and the
bucket/policy package is removed. All tests from the bucket/policy
package are moved into iam/policy and updated to work with the renamed
types.

Merge after:
* [x] https://github.com/minio/pkg/pull/79


donatello:
This is now ready. It is just a continuation of #79 where common policy code was moved into `iam/policy` and this one moves the code unique to bucket policy into the iam/policy package with types renamed to reflect that they apply to Bucket Policy only. PTAL

donatello:
After this change, `iam/policy` would be a single package exporting types for both bucket and iam policies.

donatello:
PTAL - this is ready.

krisis (approved):

harshavardhana (approved):

PR #82, opened 2023-08-05
remove duplicated code and simplify CA certs loading

harshavardhana:


jiuker (approved):

vadmeste (commented):
Is it okay if this only loads certs under specific a folder but not the system default directory ? if yes, LGTM

harshavardhana:
> Is it okay if this only loads certs under specific a folder but not the system default directory ? if yes, LGTM

it will load system and the SSL_CERT_DIR @vadmeste

vadmeste:
> On Unix systems other than macOS the environment variables SSL_CERT_FILE and SSL_CERT_DIR can be used to override the system default locations for the SSL certificate file and SSL certificate files directory, respectively. The latter can be a colon-separated list.

https://pkg.go.dev/crypto/x509

It doesn't seem it loads the default certs, what are you trying to do in the first place?

harshavardhana:
> > On Unix systems other than macOS the environment variables SSL_CERT_FILE and SSL_CERT_DIR can be used to override the system default locations for the SSL certificate file and SSL certificate files directory, respectively. The latter can be a colon-separated list.
> 
> https://pkg.go.dev/crypto/x509
> 
> It doesn't seem it loads the default certs, what are you trying to do in the first place?



> > On Unix systems other than macOS the environment variables SSL_CERT_FILE and SSL_CERT_DIR can be used to override the system default locations for the SSL certificate file and SSL certificate files directory, respectively. The latter can be a colon-separated list.
> 
> https://pkg.go.dev/crypto/x509
> 
> It doesn't seem it loads the default certs, what are you trying to do in the first place?

https://cs.opensource.google/go/go/+/refs/tags/go1.20.7:src/crypto/x509/root_unix.go;l=41

That documentation is wrong and incomplete from how the code.

The intention is to add new folders into system cert list via the environment instead of writing the entire code on our end.

certs/cert_pool_nix.go
```diff
@@ -35,52 +33,14 @@ var certDirectories = []string{
...
-	if fi.Mode()&os.ModeSymlink == 0 {
-		return false
-	}
-	target, err := os.Readlink(filepath.Join(dir, fi.Name()))
-	return err == nil && !strings.Contains(target, "/")
-}
-
 func loadSystemRoots() (*x509.CertPool, error) {
+	// Add additional ENV to load k8s CA certs.
+	os.Setenv("SSL_CERT_DIR", strings.Join(certDirectories, ":"))
```
vadmeste, line 38: Maybe you can append instead of overwriting this env variable. It is good to have this env variable configurable by the user.

harshavardhana: Yeah can do that @vadmeste

PR #85, opened 2023-11-06
basic composable sftp server boilerplate

zveinn:
This composable SFTP implementation was made in preparation for extracting the current SFTP implementation details out of [minio](github.com/minio/minio) so they can be used elsewhere.

I did not add any code that changes the basic functionality of the current SFTP boilerplate code within [minio](github.com/minio/minio). We can improve on that later if needed.

The implementation is in-line with the current FTP package we are using which makes the composition of the FTP and SFTP packages feel familiar.

# Notes
At first I was thinking about refactoring this into `github.com/minio/internal` but it felt strange to be imporint `github.com/minio` into the minwall which is where we would also be using this pkg. The second option was to create a new repository for this wrapper, but that also felt like it would be too much. So I ended up placing it here. 
If there is a better place for it then let me know!

# Usage
The SFTP pkg usage would look something like this
image

# Logger
The Logger interface could be used like this
image

### DIFF
--- a/sftp/sftp.go
+++ b/sftp/sftp.go
@@ -1,0 +1,238 @@
+// Copyright (c) 2022-2023 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package sftp
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"time"
+
+	"golang.org/x/crypto/ssh"
+)
+
+type (
+	// LogType defined various types of logs and errors
+	// that can happen within the SFTP implementation
+	LogType string
+)
+
+var (
+	// ErrMissingConnectionHandlerFunction ...
+	ErrMissingConnectionHandlerFunction = errors.New("new connection handler is not defined")
+	// ErrMissingSSHConfig ...
+	ErrMissingSSHConfig = errors.New("ssh Config is not defined")
+	// ErrMissingLoggerInterface ...
+	ErrMissingLoggerInterface = errors.New("logger interface is not defined")
+	// ErrInvalidPort ...
+	ErrInvalidPort = errors.New("port must not be 0 or bigger then 65535")
+)
+
+const (
+	// ServerStarted is logged when the SFTP server is first launched.
+	ServerStarted LogType = "server-started"
+	// ChannelNotSession is logged when the SFTP receives a request for a new channel which is NOT of type 'session'.
+	ChannelNotSession LogType = "channel-not-session"
+	// AcceptNetworkError is logged when there is an error accepting network connections within the listener.
+	AcceptNetworkError LogType = "accept-network-error"
+	// SSHKeyExchangeError is logged when there is an error performing a key exchange between the SFTP client and server.
+	SSHKeyExchangeError LogType = "ssh-key-exchange-error"
+	// AcceptChannelError is logged when there is an error while trying to accept the new request channel.
+	AcceptChannelError LogType = "accept-channel-error"
+)
+
+// Logger implements a basic logging interface
+// for the SFTP server.
+type Logger interface {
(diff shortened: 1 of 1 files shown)


sftp/sftp.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+type (
+	// LogType defined various types of logs and errors
+	// that can happen within the SFTP implementation
+	LogType string
+)
+
+const (
+	missingConnectionHandlerFunction = "New connection handler is not defined"
```
klauspost, line 37: Make these exported `var ErrMissingConnectionHandlerFunction = errors.New("new connection handler is not defined")` so callers can compare.

sftp/sftp.go
```diff
@@ -0,0 +1,228 @@
...
+	ChannelNotSession LogType = "channel-not-session"
+	// AcceptNetworkError is logged when there is an error accepting network connections within the listener.
+	AcceptNetworkError LogType = "accept-network-error"
+	// SSHKeyExchangeError is logged when there is an error performing a key exchange between the SFTP client and server.
+	SSHKeyExchangeError LogType = "ssh-key-exchange-error"
+	// AcceptChannelError is logged when there is an error while trying to accept the new request channel.
+	AcceptChannelError LogType = "accept-channel-error"
+)
+
+type logger interface {
```
klauspost, line 54:
```suggestion
type Logger interface {
```
Since this is an option - it would be nice to export it as well.

sftp/sftp.go
```diff
@@ -0,0 +1,228 @@
...
+	PublicIP  string
+	Logger    logger
+	SSHConfig *ssh.ServerConfig
+	// SSHHandshakeDeadline controls the time.Duration which ssh session
+	// have to complete their handshake. This option is not a part of the
+	// ssh.ServerConfig so we had to implement it separately.
+	SSHHandshakeDeadline time.Duration
+	// BeforeHandle will be executed before `HandleSFTPSession` and before
+	// error checking happens during the socket listener.Accept().
+	BeforeHandle func(conn net.Conn, err error) bool
```
klauspost, line 84: What does the bool signify?

sftp/sftp.go
```diff
@@ -0,0 +1,228 @@
...
+		return nil, errors.New(missingSSHConfig)
+	}
+	if options.Logger == nil {
+		return nil, errors.New(missingLoggerInterface)
+	}
+	if options.Port < 1 || options.Port > 65535 {
+		return nil, errors.New(invalidPort)
+	}
+	if options.BeforeHandle == nil {
+		options.BeforeHandle = defaultBeforeAccept
```
klauspost, line 111: Can't it just remain nil and you check before you use it?

sftp/sftp.go
```diff
@@ -0,0 +1,228 @@
...
+					tempDelay *= 2
+				}
+				if max := 1 * time.Second; tempDelay > max {
+					tempDelay = max
+				}
+				s.logger.Error(AcceptNetworkError, fmt.Errorf("error while accepting connections: %w, retrying in %s", err, tempDelay))
+				time.Sleep(tempDelay)
+				continue
+			}
+			return err
```
klauspost, line 169: It seems a bit drastic to exit the entire server handler because of an error. We would the very least have to be sure this cannot be trigged by anything on the user side.

Overall, I think it would be better to just log and move on - otherwise it is a potential DOS target.

sftp/sftp.go
zveinn: It's used every time we accept a connection, so I figured a single check would be better then multiple.

sftp/sftp.go
klauspost: Yeah. Didn't even consider the delay functionality, which basically can pretty much lock up incoming requests.

klauspost: Personally, I'd just take it all out. Log the error and move on.

zveinn: Agreed, I just did that. 

Currently I am distinguishing between Timeout() errors and non-timeout errors. With timeout errors triggering a continue and the non-timeout once triggering a full return. But honestly, i'm not sure about that either. I kind of feel like 99% of error cases here are going to merit a continue. 

I might look into this a bit more today.

zveinn: I looked into it and it seems that the Temporary functionality has not been changed, the method has just been marked as deprecated. 

I'm going to re-implement Temporary along with Timeout so that our new SFTP package is consistent with the current implementation.

sftp/sftp.go
klauspost: Without checking you are setting up the function call and calling+returning every time - since it is a function pointer it can never be skipped however if we do:

`if options.BeforeHandle != nil && !options.BeforeHandle(....) {`

... the only cost will be the nil check - which the compiler does anyway if you don't add it.

Since it is a fully predictable branch you really aren't leaving anything on the table.

zveinn: you are right, I didn't even think about that option. I'll change it right away

harshavardhana (approved):

poornas (approved):

PR #86, opened 2023-11-08
Adding Context awareness and the ability to control tcp keep-alive

zveinn:
These changes enable us to pass a context into the sftp server for a controlled shutdown. Additionally it enables optional tcp keep-alive configurations.


klauspost (approved):
LGTM

harshavardhana (approved):

PR #87, opened 2023-11-22
net.Listener does not close when context.cancelFunc is called + version bump for vauln fixes

zveinn:
Apparently the net.Listener does not respect the cancelFunc call. Instead we have to directly call .Close()

### DIFF
--- a/.github/workflows/ldap.yaml
+++ b/.github/workflows/ldap.yaml
@@ -31,7 +31,7 @@ jobs:
           LDAP_ADMIN_PASSWORD: "admin"
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x]
+        go-version: [1.21.x]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v3
--- a/.github/workflows/ldap.yaml
+++ b/.github/workflows/ldap.yaml
@@ -31,7 +31,7 @@ jobs:
           LDAP_ADMIN_PASSWORD: "admin"
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/minio/pkg/v2
 
-go 1.19
+go 1.21.1
```
harshavardhana, line 3:
```suggestion
go 1.20
```

.github/workflows/vulncheck.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ 1.19.x, 1.20.x ]
+        go-version: [ 1.21.x ]
```
harshavardhana, line 17:
```suggestion
        go-version: [ 1.20.x, 1.21.x ]
```

.github/workflows/ldap.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
           LDAP_ADMIN_PASSWORD: "admin"
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x]
+        go-version: [1.21.x]
```
harshavardhana, line 34:
```suggestion
        go-version: [ 1.20.x, 1.21.x ]
```

harshavardhana (changes requested):

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #90, opened 2023-12-13
fix govulncheck errors

anjalshireesh:
- Change go version from 1.20 to 1.21
- Upgrade github.com/lestrrat-go/jwx from v1.2.26 to v1.2.27


go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/minio/pkg/v2
 
-go 1.20
+go 1.21
```
harshavardhana, line 3:
```suggestion
go 1.19
```

harshavardhana (approved):

harshavardhana:
we need to keep the `go.mod` 1.19 so that we can always build with go1.19 @anjalshireesh

PR #91, opened 2023-12-13
Adding GetObjectAttributes S3 action to our policies list

zveinn:


klauspost (approved):
LGTM

klauspost:
Seems like [CI Go version](https://github.com/minio/pkg/blob/main/.github/workflows/vulncheck.yml#L17) needs a bump along with github.com/lestrrat-go/jwx

zveinn:
> Seems like [CI Go version](https://github.com/minio/pkg/blob/main/.github/workflows/vulncheck.yml#L17) needs a bump along with github.com/lestrrat-go/jwx

yea, there is another PR to fix the tests. If they are still failing when I'm about to be done with this I'll grab that one too.

harshavardhana (approved):

harshavardhana (changes requested):
Actually this is not sufficient @zveinn you need to add this into supported actions list and also add relevant headers that are valid. 

Look for other actions where all they are referenced check `GetObjectAction`

zveinn:
> Actually this is not sufficient @zveinn you need to add this into supported actions list and also add relevant headers that are valid.
> 
> Look for other actions where all they are referenced check `GetObjectAction`

Will do

policy/action.go
```diff
@@ -372,7 +374,9 @@ func createActionConditionKeyMap() ActionConditionKeyMap {
 				condition.ExistingObjectTag.ToKey(),
 			}, commonKeys...)...),
 
-		HeadBucketAction: condition.NewKeySet(commonKeys...),
+		HeadBucketAction:                 condition.NewKeySet(commonKeys...),
+		GetObjectAttributesAction:        condition.NewKeySet(commonKeys...),
```
harshavardhana, line 378: https://docs.aws.amazon.com/AmazonS3/latest/userguide/list_amazons3.html Actually this supports tags as well so you need to add 

```
		GetObjectAttributesAction: condition.NewKeySet(
			append([]condition.Key{
				condition.S3VersionID.ToKey(),
				condition.ExistingObjectTag.ToKey(),
			}, commonKeys...)...),
```

Not just commonKeys.

harshavardhana (changes requested):

policy/action.go
zveinn: GetObjectAttributes did not require the versionID, the GetObjectVersionAttributes was the one. 

But they've been added now.

klauspost (approved):

harshavardhana (approved):

PR #99, opened 2024-02-29
Take accepted plans as input to validateEnterpriseLicense

anjalshireesh:
There are going to be multiple variants of ENTERPRISE license. So add ability to pass a list of accepted plans to the validation function.

Also read the optional new field `trial` from the license and use it in the validation (trial licenses don't get grace period after expiry)

### DIFF
--- a/licverifier/verifier.go
+++ b/licverifier/verifier.go
@@ -49,6 +49,7 @@ type LicenseInfo struct {
 	IssuedAt        time.Time // Time of license issue
 	ExpiresAt       time.Time // Time of license expiry
 	APIKey          string    // Subnet account API Key
+	IsTrial         bool      // Is this a TRIAL license?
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

harshavardhana:
> There are going to be multiple variants of ENTERPRISE license. So add ability to pass a list of accepted plans to the validation function.

is this still true?

anjalshireesh:
> > There are going to be multiple variants of ENTERPRISE license. So add ability to pass a list of accepted plans to the validation function.
> 
> is this still true?

@harshavardhana yes, discussed and decided yesterday. New plans will be `ENTERPRISE-LITE` and `ENTERPRISE-PLUS`

kanagarajkm (approved):

PR #100, opened 2024-03-10
Fix handling of empty values in `env.LookupEnv` and `env.Get`

twelho:
With the changes from https://github.com/minio/pkg/commit/716cd4d05b704a00ae42bdacc383134f7763e2f0, `env.LookupEnv` and `env.Get` have no longer been able to distinguish between environment variables that are not set and variables that are set, but have an empty string as the value. This contradicts the descriptions of both, which state that an empty value should still be respected and correctly handled. Additionally, `env.IsSet` also incorrectly considers set-but-empty environment variables as unset.

Importantly, this uncovered a bug in MinIO, where setting `MINIO_COMPRESSION_EXTENSIONS=` and `MINIO_COMPRESSION_MIME_TYPES=` causes the environment variables to be ignored, even though this is the intended way to [enable compression for all compressible objects](https://min.io/docs/minio/linux/administration/object-management/data-compression.html#compress-all-compressible-objects) when using [environment variables](https://min.io/docs/minio/linux/reference/minio-server/settings/core.html#id12), leading to the default compression extension and MIME type filters being applied. The fix for MinIO can be found in PR TODO.

This fix necessarily changes the semantics of `env.LookupEnv` slightly to expose the underlying existence check boolean (again), and needs minor changes on MinIO's side (in PR TODO), but at least https://github.com/minio/console and https://github.com/minio/mc do not need any changes in reponse.

For `env.IsSet`, I have also added a check that considers all environment variables as unset if environment lookup is disabled, which better matches the intention of the function. However, I can remove this check if looking up the set status should be possible even when lookups are disabled.

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -55,7 +55,15 @@ func SetEnvOn() {
 
 // IsSet returns if the given env key is set.
 func IsSet(key string) bool {
-	return Get(key, "") != ""
+	privateMutex.RLock()
+	ok := envOff
+	privateMutex.RUnlock()
+	if !ok {
+		_, ok, _, _, _ = LookupEnv(key)
+		return ok
+	}
+	// Env lookup is off
+	return false
 }
 
 // Get retrieves the value of the environment variable named
@@ -67,8 +75,8 @@ func Get(key, defaultValue string) string {
 	ok := envOff
 	privateMutex.RUnlock()
 	if !ok {
-		v, _, _, _ := LookupEnv(key)
-		if v != "" {
+		v, ok, _, _, _ := LookupEnv(key)
+		if ok {
 			return strings.TrimSpace(v)
 		}
 	}
--- a/env/web_env.go
+++ b/env/web_env.go
@@ -173,7 +173,7 @@ func Environ() []string {
 // Additionally if the input is env://username:password@remote:port/
 // to fetch ENV values for the env value from a remote server.
 // In this case, it also returns the credentials username and password
-func LookupEnv(key string) (string, string, string, error) {
+func LookupEnv(key string) (string, bool, string, string, error) {
 	v, ok := os.LookupEnv(key)
 	if ok && strings.HasPrefix(v, webEnvScheme) {
 		// If env value starts with `env*://`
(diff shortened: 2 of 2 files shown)


harshavardhana:
While the description is not updated, this was the intended behavior. We must update the description instead of breaking it. 

The reason we need to know that is because we must treat `empty` value with ENV set as incorrect behavior, you must not have an ENV with empty values. If you want to set empty value to ENV you must not have it in the first place. 

I will update the description of the intended behavior and match it what we want.

harshavardhana:
> Importantly, this uncovered a bug in MinIO, where setting `MINIO_COMPRESSION_EXTENSIONS=` and `MINIO_COMPRESSION_MIME_TYPES=` causes the environment variables to be ignored, even though this is the intended way to [enable compression for all compressible objects](https://min.io/docs/minio/linux/administration/object-management/data-compression.html#compress-all-compressible-objects) when using [environment variables](https://min.io/docs/minio/linux/reference/minio-server/settings/core.html#id12),

I think that is wrong from a documentation point of view. We must have a value there instead of using `empty` values. We must say `*` there to apply it for all objects and all extensions.

twelho:
Please take a look at https://github.com/minio/minio/pull/19233, which is why this is needed. If you have an alternative suggestion of how to fix this issue, I would be happy to implement it.

harshavardhana:
> Please take a look at [minio/minio#19233](https://github.com/minio/minio/pull/19233), which is why this is needed. If you have an alternative suggestion of how to fix this issue, I would be happy to implement it.

Yes that PR is wrong. We should not support empty values in that manner.

- Empty means inherit default values, if any, that functionality mandates. 
- Since this is a regex input, take regex input instead of taking empty as "all-encompassing universal value"
-

twelho:
Only the MIME types support regex, the extensions rely on `hasStringSuffixInSlice` which does not. Thus, for every object, extensions are always checked regardless before short-circuiting with the MIME regex, which seems inelegant.

harshavardhana:
> Only the MIME types support regex, the extensions rely on `hasStringSuffixInSlice` which does not. Thus, for every object, extensions are always checked regardless before short-circuiting with the MIME regex, which seems inelegant.

The correct fix is honor `*` properly. Not by allowing empty values in ENV and then assuming that its a "set" value to something called as `*`

twelho:
Should I implement that as the fix instead to avoid https://github.com/minio/minio/pull/19233#issuecomment-1987364579?

twelho:
Implemented in https://github.com/minio/minio/pull/19234.

PR #101, opened 2024-03-12
Exit if license expired

anjalshireesh:
If the license has expired (and grace period also over where applicable), then exit from the program.

To achieve this, the function `ValidateEnterpriseLicense` which is called at application startup will also schedule the next license check after successful validation.

If the license is currently not expired, schedule to check again just after expiry. If it is expired but within grace period, check every day.

### DIFF
--- a/subnet/license.go
+++ b/subnet/license.go
@@ -196,7 +196,7 @@ func getDurationForNextLicenseCheck(li *licverifier.LicenseInfo) time.Duration {
 	return time.Until(li.ExpiresAt.Add(time.Second))
 }
 
-func (lv *LicenseValidator) scheduleNextLicenseCheck(li *licverifier.LicenseInfo, acceptedPlans []string) error {
+func (lv *LicenseValidator) scheduleNextLicenseCheck(li *licverifier.LicenseInfo, acceptedPlans []string, licExpiredChan chan<- string) {
 	duration := getDurationForNextLicenseCheck(li)
 	timer := time.NewTimer(duration)
 	defer timer.Stop()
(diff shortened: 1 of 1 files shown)


subnet/license.go
```diff
@@ -170,6 +172,36 @@ func (lv *LicenseValidator) ValidateLicense() (*licverifier.LicenseInfo, error)
...
+	timer := time.NewTimer(duration)
+	defer timer.Stop()
+
+	ctxt := context.Background()
+	for {
+		select {
+		case <-timer.C:
+			li, err := lv.ValidateEnterpriseLicense(acceptedPlans)
+			if err != nil {
+				fmt.Println("license validation failed:", err.Error())
```
kannappanr, line 195: We need not print here. Application that uses this library, might not be expecting this sort of printing in the library

harshavardhana, line 195: Just return error, no printing needed and do not exit here.

subnet/license.go
```diff
@@ -188,11 +220,22 @@ func (lv *LicenseValidator) ValidateEnterpriseLicense(acceptedPlans []string) (*
...
 
-	if li.ExpiresAt.Before(time.Now()) && li.IsTrial {
-		return nil, fmt.Errorf("trial license has expired on %v", li.ExpiresAt)
+	if li.ExpiresAt.Before(time.Now()) {
+		if li.IsTrial || li.Plan == "TRIAL" {
+			// no grace period for trial
+			return nil, fmt.Errorf("trial license has expired on %v", li.ExpiresAt)
+		}
+		// expired, within grace period. print warning.
+		fmt.Printf("license has expired on %v, renew immediately to avoid outage.", li.ExpiresAt)
```
harshavardhana, line 232:
```suggestion
		fmt.Printf("license has expired on %v, renew immediately to avoid outage.", li.ExpiresAt)
```

All this logging must be avoided here, the library must just generates errors or reports for the caller to get the right set of information to show.

It should be printing this on its own, this can cause problems for logging systems under each of these projects.

harshavardhana (changes requested):

subnet/license.go
anjalshireesh: @harshavardhana exiting directly from here because this is happening in background, so can't just return error. And we want enterprise tools to exit when license has expired. Do you want me to implement async communication with the app (using channels?) and let the application code decide how to exit on receiving license expired message?

anjalshireesh: Thanks, done

anjalshireesh: @harshavardhana I've modified the code to return the error through a license expiry channel. PTAL.

subnet/license.go
anjalshireesh: Removed all logging.

kannappanr (approved):

harshavardhana (approved):

PR #102, opened 2024-03-18
resources are allowed to be empty

harshavardhana:


harshavardhana:
admin and kms actions from get-go have never had resources, we were lucky enough somehow this code was never triggered with jsoniter on am empty resourceSet.

However a migration to gojson revealed this bug in https://github.com/minio/minio/pull/19281

PR #104, opened 2024-03-27
ldap: normalize DNs before returning lookup result

donatello:


ldap/ldap.go
```diff
@@ -250,7 +254,21 @@ func getGroups(conn *ldap.Conn, sreq *ldap.SearchRequest) ([]string, error) {
...
+		}
+		groups = append(groups, normalizedDN)
 	}
 	return groups, nil
 }
+
+// NormalizeDN normalizes the DN. The ldap library here mainly lowercases the
+// attribute type names in the DN.
+func NormalizeDN(dn string) (string, error) {
+	parsedDN, err := ldap.ParseDN(dn)
```
harshavardhana, line 269: would this be a breaking change?

donatello: No it should not be. OpenLDAP already lowercases and returns like "DC=Min,DC=io" becomes "dc=Min,dc=io". We are only additionally doing it here, in case other servers like AD don't. The idea is to mostly assume the server returns a cononical value for DNs - OpenLDAP server already performs unicode normalization etc. It's hard to do all that in our code, because the LDAP spec is quite involved.

harshavardhana (approved):

PR #105, opened 2024-03-27
ldap: Include server name is tls config

donatello:
This change adds the server name to the TLS config in the LDAP client connection object. When upgrading a connection using STARTTLS it seems to be required to verify the server certificate.

Additionally, stop using deprecated function in the go-ldap library and switch to supported ones.

### DIFF
--- a/ldap/ldap.go
+++ b/ldap/ldap.go
@@ -68,30 +68,37 @@ func (l *Config) Clone() (cloned Config) {
 	return cloned
 }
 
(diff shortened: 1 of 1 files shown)


donatello:
Marking this as draft until some testing with tls enabled openldap.

harshavardhana:
Is this still needed @donatello ?

donatello:
No I was waiting on some user input on this but they never came back. I don't think it's needed.

PR #107, opened 2024-04-12
fix:not startTls connect use tls.

jiuker:
fix:not startTls connect use tls.

fix https://github.com/minio/operator/issues/2048


shtripat (approved):
lgtm

jiuker:
on purpose.

PR #109, opened 2024-04-17
cond: Support NumericGreaterThanIfExists condition

vadmeste:
Add support of NumericGreaterThanIfExists condition. 

NumericGreaterThanIfExists means the condition returns true if the condition is not passed


harshavardhana (approved):

harshavardhana (changes requested):
Please add a test case

vadmeste:
> Please add a test case

Done

harshavardhana (approved):

PR #112, opened 2024-06-03
move license verification to `license` package

aead:
This commit adds the `license` package and removes the `licverifier` and `subnet` packages. The new
`license` package exposes a simpler API for verifying licenses that is better suited for conditional compilation.

Ideally, license verification should only happen for release builds. Hence, license verification code can be guarded by a build directive.
```
//go:build !dev
```


harshavardhana (approved):

licenses/license.go
```diff
@@ -0,0 +1,267 @@
...
+
+// Verify checks whether a valid license is provided.
+//
+// Therfore, it first searches for a license in the following order:
+//  1. File referenced by MINIO_LICENSE env var, if any.
+//  2. A "./minio.license" file in the current working directory.
+//  3. The $HOME/"." + os.Args[0]/minio.license file if there are os.Args.
+//
+// If no license is present, Verify returns an error.
+// The license must also be issued after time.Now and,
```
anjalshireesh, line 176:
```suggestion
// The license must also be issued before time.Now and,
```

licenses/license.go
```diff
@@ -0,0 +1,267 @@
...
+	l, err := Parse(string(license), jwt.WithKeySet(PublicKey()), jwt.UseDefaultKey(true), jwt.WithValidate(true), jwt.WithAcceptableSkew(licenseSkew))
+	if err != nil {
+		return fail(err)
+	}
+	if l.IssuedAt.After(time.Now()) {
+		return fail(errors.New("license: license is not yet valid"))
+	}
+	if l.Plan != planEnterpriseLite && l.Plan != planEnterprisePlus {
+		return fail(fmt.Errorf("license: either %s or %s plan required", planEnterpriseLite, planEnterprisePlus))
+	}
```
anjalshireesh, line 218: In case of mineos , the legacy plans also need to be supported (`"TRIAL", "STANDARD", "ENTERPRISE"`)

licenses/license.go
```diff
@@ -0,0 +1,267 @@
...
+		IssuedAt:     token.IssuedAt(),
+		ExpiresAt:    token.Expiration(),
+		NotAfter:     token.Expiration(),
+		Trial:        isTrial,
+	}, nil
+}
+
+// Verify checks whether a valid license is provided.
+//
+// Therfore, it first searches for a license in the following order:
```
anjalshireesh, line 170:
```suggestion
// Therefore, it first searches for a license in the following order:
```

licenses/license.go
```diff
@@ -0,0 +1,267 @@
...
+//	func init() {
+//		l, err := license.Verify()
+//		if err != nil {
+//			fmt.Fprintln(os.Stderr, "Error:", err)
+//			os.Exit(1)
+//		}
+//		if time.Now().After(l.ExpiresAt) {
+//			fmt.Fprintf(os.Stderr, "Warning: license expired on %v! Renew and restart immediately to avoid outage.\n", l.ExpiresAt)
+//		}
+//		if time.Now().After(l.NotAfter) {
```
anjalshireesh, line 31: Can this be part of `Verify()` itself?

licenses/license.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/lestrrat-go/jwx/jwt"
+
+	"time"
+)
+
+const publicKey = `-----BEGIN PUBLIC KEY-----
+MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEaK31xujr6/rZ7ZfXZh3SlwovjC+X8wGq
+qkltaKyTLRENd4w3IRktYYCRgzpDLPn/nrf7snV/ERO5qcI7fkEES34IVEr+2Uff
+JkO2PfyyAYEO/5dBlPh1Undu9WQl6J7B
+-----END PUBLIC KEY-----`
```
anjalshireesh, line 72: We may need some mechanism to get new key in case it changes on SUBNET side.

aead: You would only ever want to change a key if it is compromised. If that is the case, the new key should be added here and this one should be removed (after a grace period). It seems reasonable to expect people to upgrade within this timeframe.

Dynamically fetching the key allows someone intercepting the network traffic to replace it and craft their own custom licenses.

licenses/license.go
aead: Really? So the behavior for a e.g. a "STANDARD" plan user would be that he can run mineos but not e.g. minwall. Are there users doing this @harshavardhana ?

If so we can change it to `license.Verify(plans ...string)`, I guess.

licenses/license.go
aead: Actually, yes. This would be only required if `license.Parse` is used for using a different license skew time (not 30 days).

licenses/license.go
harshavardhana: We have to provide fallbacks when that happens so we can have two keys at a time. The one that we always use , and the one that can be injected freshly into the implementation.

aead: > All products doing license verification must immediately be released with the new key (as existing versions will fail verification of newly generated licenses)

All products must be released with a single dep change: Upgraded `minio/pkg`. This can be done as a regular release or (if not possible at the moment) from the latest stable with a backport. Also, there is "immediate" action required since if the key is compromised, any existing binary / container image can still be used. This more on us then the users/customers.

> Any existing customer whose license has expired (and grace period also over) must upgrade to the new version after getting their license extended/renewed (as the new license will be generated with new key)

Technically yes. However, in cases where this is not possible we could make an exception and provide a license signed with the compromised key. 

> If we don't want to enforce upgrade in these situations, one option could be to allow the customer to override the public key by setting the new one in an env or standard file path.

This is even worse then allowing the key to be fetched from the network. If you can provide your own public key for verifying the binary, then you just create a private/public key pair, issue an arbitrary license with the private key and provide the public key you just generated.

aead: Let's assume our private signing key `K1` gets compromised. This means someone can issue arbitrary licenses that any exisiting release (up to this day) considers valid. Would this be different if we had e.g. a 2nd public key in the binary? No, but customers would not be required to update when they renew their license.

If `K1` gets compromised, we have to make new releases of all binaries that require a license since existing binaries always accept licenses issued by `K1`. We have no proper way of telling these binaries that they should no longer accept licenses issued by `K1`. We have two options now:
 - Build a binary that only accepts licenses issued by our new key `K2`. This requires that we give customers a new license.
 - Build a binary that accepts licenses issued by `K1` or `K2`. In this case, the customer does not have to update it's license.

Including two public keys (`P1` and `P2`) in the binary **only** allows customers to keep their current version of the binary but use a new license - issued by `K2`.

***

> Existing binary cannot be used by new customers because their license would have been created with the new key and hence would fail verification by existing binary.

This is not clear to me. If `K1` is compromised and a customer wants to use a binary that only accepts licenses issued by `K1` we still can issue a license using `K1` for this customer. Basically:
If customer uses release before `K1` got compromised, issue license with `K1`. Otherwise issue license using `K2`.

aead: Let's assume our private signing key `K1` gets compromised.
Now, someone can issue arbitrary licenses that any existing
binary considers valid. Hence, we switch to `K2` and start
issue licenses with it from now on.

Now, there are:
 - Users running existing binaries with licenses issued by `K1`.
   This is fine.
 - Users running existing binaries with licenses issued by `K2`.
   This might be a problem.
 - Users running new binaries with licenses issued by `K1`.
   This might be a problem.
 - Users running new binaries with licenses issued by `K2`.
   This is fine.

When a user has a license issued by `K1` and upgrades its binary
to a version requiring `K2` then the user has to update its license.
This is always the case.

When a user has a license isssued by `K2` and tries to run a
an existing binary then `P2` (public key) must be already in
the binary or it will not work. Otherwise, the user is forced to
upgrade.

Hence, putting a 2nd public key (`P2`) into the binaries now
allows users to update their license without upgrading their
binaries. However, this requires that `K2` is stored separatley
from `K1` (e.g. offline).

An alternative approach would be to issue the license based
on the user's binary version. If the binary still requires a
license issued by `K1` we can issue one. Nothing prevents
that.

In any case, users that upgrade will require a new license eventually.

aead: >Existing binary cannot be used by new customers because their license would have been created with the new key and hence would fail verification by existing binary.

What I'm trying to say is that a customer does not have to upgrade immediately. We only start issuing new licenses when releases are ready - not as soon as we notice that the key got compromised. Once new releases are ready, new customers will use them most probably. (Special cases can be taken care of). There is no point in changing the key as long as no releases that can verify it are available.

PR #120, opened 2024-07-09
feat: support xtime duration

jiuker:
feat: support xtime duration


xtime/time_unmarshal_test.go
```diff
@@ -0,0 +1,41 @@
+package xtime
+
```
harshavardhana, line 2: Add missing copyright

harshavardhana (changes requested):

PR #121, opened 2024-07-12
Allow a KMS Action to specify keys in the Resources of a policy

marktheunissen:


policy/resource.go
```diff
@@ -44,8 +89,18 @@ func (r Resource) isObjectPattern() bool {
...
-		return false
+	if r.isAWSS3() {
+		if strings.HasPrefix(r.Pattern, "/") {
+			return false
+		}
+	}
+	if r.isKMS() {
+		// TODO: Copied from KES repo, confirm that it is generally applicable.
+		if strings.IndexFunc(r.Pattern, func(c rune) bool {
+			return c == '/' || c == '\\' || c == '.'
```
marktheunissen, line 100: Is this a good rule for validating key names? Copied from KES.

policy/resource.go
```diff
@@ -26,12 +26,57 @@ import (
...
+const (
+	// ResourceARNPrefix - resource S3 ARN prefix as per AWS S3 specification.
+	ResourceARNPrefix = "arn:aws:s3:::"
+
+	// ResourceARNKMSPrefix is for KMS key resources. MinIO specific API.
+	ResourceARNKMSPrefix = "arn:minio:kms:::"
+)
+
+// ARNPrefixType - ARN prefix type
+type ARNPrefixType uint8
```
harshavardhana, line 38:
```suggestion
type ResourceARNType int32

const (
    ResourceARNS3 ResourceARNType = iota
    ResourceARNKMS
    
    unknownARN // for detecting errors
)
```

policy/resource.go
```diff
@@ -44,8 +89,18 @@ func (r Resource) isObjectPattern() bool {
...
 func (r Resource) IsValid() bool {
-	if strings.HasPrefix(r.Pattern, "/") {
-		return false
+	if r.isAWSS3() {
+		if strings.HasPrefix(r.Pattern, "/") {
+			return false
+		}
+	}
+	if r.isKMS() {
+		// TODO: Copied from KES repo, confirm that it is generally applicable.
```
harshavardhana, line 98:
```suggestion
```

yes it is acceptable

policy/resource.go
harshavardhana: 👍🏼

policy/resource.go
```diff
@@ -136,23 +191,38 @@ func (r Resource) ValidateBucket(bucketName string) error {
...
-	if strings.HasPrefix(pattern, "/") {
+	if strings.HasPrefix(r.Pattern, "/") {
 		return Resource{}, Errorf("invalid resource '%v' - starts with '/' will not match a bucket", s)
 	}
 
+	return r, nil
+}
+
+// NewResourceAWSS3 - creates new resource with the default ARN type of AWS S3.
+func NewResourceAWSS3(pattern string) Resource {
```
harshavardhana, line 215:
```suggestion
func NewResourceS3(pattern string) Resource {
```

policy/statement.go
```diff
@@ -129,7 +141,13 @@ func (statement Statement) isValid() error {
...
 
 	if statement.isKMS() {
-		return statement.Actions.ValidateKMS()
+		if err := statement.Actions.ValidateKMS(); err != nil {
+			return err
+		}
+		if err := statement.Resources.ValidateKMS(); err != nil {
+			return err
+		}
+		return nil
```
harshavardhana, line 150:
```suggestion
		return statement.Resources.ValidateKMS()
```

harshavardhana (changes requested):

marktheunissen:
Should be good to go, all feedback implemented 👍

policy/resourceset.go
```diff
@@ -150,9 +150,26 @@ func (resourceSet *ResourceSet) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// Validate - validates ResourceSet.
-func (resourceSet ResourceSet) Validate() error {
+// ValidateAWSS3 - validates ResourceSet is AWS S3.
+func (resourceSet ResourceSet) ValidateAWSS3() error {
```
harshavardhana, line 154: don't name any thing AWSS3 its just S3 in all those situations.

policy/resourceset.go
```diff
@@ -150,9 +150,26 @@ func (resourceSet *ResourceSet) UnmarshalJSON(data []byte) error {
...
 	for resource := range resourceSet {
+		if !resource.isAWSS3() {
+			return Errorf("type of resource is not AWS S3 '%v'", resource)
```
harshavardhana, line 157:
```suggestion
			return Errorf("resource '%v' type is not S3", resource)
```

harshavardhana (changes requested):

policy/resource.go
```diff
@@ -26,12 +26,59 @@ import (
...
+type ResourceARNType uint32
+
+const (
+	// ResourceARNS3 is the ARN prefix type for AWS S3 resources.
+	ResourceARNS3 ResourceARNType = iota + 1
+
+	// ResourceARNKMS is the ARN prefix type for MinIO KMS resources.
+	ResourceARNKMS
+
+	unknownARN // for detecting errors
```
klauspost, line 47: Move this to value 0 - then it also catches the zero value. You can use `endARN` here, if you need the size anywhere.

That way `Resource{} ==Resource{Type: unknownARN}`, which seems reasonable.

policy/resource.go
```diff
@@ -136,23 +195,38 @@ func (r Resource) ValidateBucket(bucketName string) error {
...
+	switch {
+	case strings.HasPrefix(s, ResourceARNPrefix):
+		r.Pattern = strings.TrimPrefix(s, ResourceARNPrefix)
+		r.Type = ResourceARNS3
+
+	case strings.HasPrefix(s, ResourceARNKMSPrefix):
+		r.Pattern = strings.TrimPrefix(s, ResourceARNKMSPrefix)
+		r.Type = ResourceARNKMS
+	default:
+		return r, Errorf("invalid resource '%v'", s)
```
klauspost, line 208: Simpler and not something we need to maintain:

```suggestion
	for k, v := range ARNPrefixToType {
		if rem, ok := strings.CutPrefix(s, k); ok {
			r.Type = v
			r.Pattern = rem
			break
		}
	}
	if r.Type == unknownARN {
		return r, Errorf("invalid resource '%v'", s)
```

policy/resource.go
```diff
@@ -136,23 +195,38 @@ func (r Resource) ValidateBucket(bucketName string) error {
...
-		return Resource{}, Errorf("invalid resource '%v' - starts with '/' will not match a bucket", s)
+	if strings.HasPrefix(r.Pattern, "/") {
+		return r, Errorf("invalid resource '%v' - starts with '/' will not match a bucket", s)
 	}
 
+	return r, nil
+}
+
+// NewResourceS3 - creates new resource with the default ARN type of AWS S3.
+func NewResourceS3(pattern string) Resource {
```
klauspost, line 219: Do we need to break exported API here? Can't we just keep the name. After all the comment says what it does.

policy/resource.go
```diff
@@ -136,23 +195,38 @@ func (r Resource) ValidateBucket(bucketName string) error {
...
 		Pattern: pattern,
-	}, nil
+		Type:    ResourceARNS3,
+	}
 }
 
-// NewResource - creates new resource.
-func NewResource(pattern string) Resource {
+// NewResourceKMS - creates new resource with type KMS
+func NewResourceKMS(pattern string) Resource {
```
klauspost, line 227: Very nitpicky: I would say `NewKMSResource` is a bit more natural in terms of language. It does look sillier with all the upper case letters. Feel free to ignore.

klauspost (commented):
Just some minor tweaks.

policy/resource.go
marktheunissen: Sounds good, but I don't think we need the size? If I add it, I'll need to use it somewhere or disable linting on that line, as the linter complains that it's unused.

policy/resource.go
marktheunissen: Sure, I have changed it back.

policy/resource.go
marktheunissen: I've moved to zero value but have not added endARN, LMK if I should and how you would like it used.

harshavardhana (approved):

klauspost (approved):

PR #123, opened 2024-07-25
Supporting code for bucket-level CORS configuration settings

marktheunissen:
Adds CORS rule parsing and validation. Not sure if this should be in pkg or perhaps just straight into `minio`.

All related PRs:
- https://github.com/minio/minio/pull/20150
- https://github.com/minio/pkg/pull/123
- https://github.com/minio/madmin-go/pull/290
- https://github.com/minio/minio-go/pull/1987


cors/cors.go
```diff
@@ -0,0 +1,174 @@
...
+	ExposeHeader  []string `xml:"ExposeHeader,omitempty"`
+	ID            string   `xml:"ID,omitempty"`
+	MaxAgeSeconds int      `xml:"MaxAgeSeconds,omitempty"`
+}
+
+// Validate checks the CORS configuration is valid. This has been implemented to return errors that can be transformed
+// to match the S3 API externally, while being slightly more informative internally using wrapping.
+// Validate copies S3 behavior, and validates one rule at a time, erroring on the first invalid one found.
+func (c *Config) Validate() error {
+	allowedRuleMethods := []string{http.MethodGet, http.MethodPut, http.MethodPost, http.MethodDelete, http.MethodHead}
```
klauspost, line 54: Consider: 
```
var allowedCORSRuleMethods = map[string]bool{http.MethodGet: true, http.MethodPut: true, http.MethodPost:true, http.MethodDelete:true, http.MethodHead: true}
```
... as a package level var. Avoids the slice being created on every call and `slices.Contains(allowedRuleMethods, method)` can be `allowedCORSRuleMethods[method]`, which is just as readable.

cors/cors.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	return &c, nil
+}
+
+// ToXML marshals the CORS configuration to XML.
+func (c *Config) ToXML() ([]byte, error) {
+	if c.XMLNS == "" {
+		c.XMLNS = defaultXMLNS
+	}
```
klauspost, line 168: It seems a bit strange that ToXML is mutating the value. Maybe change the signature to `func (c Config) ToXML()`, so you aren't actually mutating the caller?

cors/errors.go
```diff
@@ -0,0 +1,65 @@
...
+
+package cors
+
+import "fmt"
+
+// ErrTooManyRules is returned when the number of CORS rules exceeds the allowed limit.
+type ErrTooManyRules struct{}
+
+func (e ErrTooManyRules) Error() string {
+	return "The number of CORS rules should not exceed allowed limit of 100 rules."
```
klauspost, line 26:
```suggestion
	return "The number of CORS rules may not exceed allowed limit of 100 rules."
```

klauspost (commented):
Just some very minor stuff. Looks very good!

cors/cors.go
marktheunissen: Good tip, thanks.

cors/errors.go
marktheunissen: This string is copied verbatim from the S3 error response when > 100 rules are given. 

`<Error><Code>InvalidRequest</Code><Message>The number of CORS rules should not exceed allowed limit of 100 rules.</Message><RequestId>XQMDEJE16GGJF7YG</RequestId><HostId>kucsoXpMyNECVKHGVNtC7mfRiqiKTZyLbYie5vn4L3bdyfamWojfDINCnwcplD+yj0FB5KD+mGNbvVSEbFUjDfcUIRl8Eo11</HostId></Error>`

I can change it here but would probably keep the user-facing error consistent with S3.

harshavardhana: Add a comment and FIXME: to update this in future when AWS fixes it on their end.

harshavardhana (approved):

harshavardhana (approved):

policy/action.go
```diff
@@ -44,6 +44,9 @@ const (
 	// DeleteBucketPolicyAction - DeleteBucketPolicy Rest API action.
 	DeleteBucketPolicyAction = "s3:DeleteBucketPolicy"
 
+	// DeleteBucketCorsAction - DeleteBucketCors Rest API action.
+	DeleteBucketCorsAction = "s3:DeleteBucketCors"
```
harshavardhana, line 48: Ah one thing missing you need to add these actions to "AllActions" as well as actions with Conditions list

marktheunissen: Will do

PR #124, opened 2024-08-05
Include the new CORS actions in supportedActions

marktheunissen:
> Ah one thing missing you need to add these actions to "AllActions" as well as actions with Conditions list

@harshavardhana The other actions that I was following as an example: (`GetBucketPolicyAction`, `GetBucketVersioningAction`, etc) only appear in `AllActions` list. 

For the "actions with Conditions list", is that `ActionConditionKeyMap`? Are the missing lines something like this:

```
DeleteBucketCorsAction: condition.NewKeySet(commonKeys...),
GetBucketCorsAction:  condition.NewKeySet(commonKeys...),
PutBucketCorsAction:  condition.NewKeySet(commonKeys...),
```

If so, should I add the missing `GetBucketPolicyAction`, etc?


harshavardhana:
> > Ah one thing missing you need to add these actions to "AllActions" as well as actions with Conditions list
> 
> @harshavardhana The other actions that I was following as an example: (`GetBucketPolicyAction`, `GetBucketVersioningAction`, etc) only appear in `AllActions` list.
> 
> For the "actions with Conditions list", is that `ActionConditionKeyMap`? Are the missing lines something like this:
> 
> ```
> DeleteBucketCorsAction: condition.NewKeySet(commonKeys...),
> GetBucketCorsAction:  condition.NewKeySet(commonKeys...),
> PutBucketCorsAction:  condition.NewKeySet(commonKeys...),
> ```
> 
> If so, should I add the missing `GetBucketPolicyAction`, etc?

Please add anything that is missing

harshavardhana (approved):

marktheunissen:
Ok I think it's all good, since `CommonKeys` will apply by default in `ActionConditionKeyMap.Lookup()`. It makes no difference if we add them explicitly or not to the map which is probably why the others are not there.

harshavardhana (approved):

PR #126, opened 2024-08-06
Use default certificate if no certificates match

ramondeklein:
The default Go TLS implementation always returns the first certificate (that's the "default" or "fallback" certificate) if none of the certificates match ([source](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/crypto/tls/common.go;l=1162)). Although that may be an issue for the client, the TLS handshake can continue. If no certificate is returned at all, the TLS handshake is aborted and the client gets a TLS failure.

MinIO always returns the default certificate if there is only one certificate (source) or when no server-name is set. If multiple certificates are added and none of them matches then `nil` is returned. Although the function returns an error, this error isn't visible in the console and/or `mc admin trace`.

This PR changes the behavior, so it always returns the default certificate when none of the certificates match (just like the default Go TLS implementation). Although this may be considered less secure (it provides the server's FQDN when hitting the server with an arbitrary server name), this information is already provided when the attacker hits the server without SNI.

### DIFF
--- a/certs/certs_test.go
+++ b/certs/certs_test.go
@@ -108,3 +108,33 @@ func TestValidPairAfterWrite(t *testing.T) {
 		t.Error("client certificate doesn't match expected certificate")
 	}
 }
+
+func TestNonMatchingCertificate(t *testing.T) {
+	ctx, cancelFn := context.WithCancel(context.Background())
+	defer cancelFn()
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

klauspost (approved):
LGTM, but not an expert on this.

PR #135, opened 2024-10-14
Move Batch job structs from minio/cmd

SSushmitha8:
Move Batch job structs from minio/cmd to pkg, to use them in aistor yaml marshaling


batch/batch.go
```diff
@@ -0,0 +1,213 @@
...
+import (
+	"context"
+	"time"
+
+	miniogo "github.com/minio/minio-go/v7"
+	"github.com/minio/pkg/v3/xtime"
+)
+
+// BatchJobRequest to start batch job
+type BatchJobRequest struct {
```
harshavardhana, line 29: there is no reason to move this to a sub-folder? what is the reason?

harshavardhana (changes requested):

batch/batch.go
harshavardhana: oh I see this is needed for external usage? the best place for this is in `madmin-go` not in this project @SSushmitha8

SSushmitha8: I want both minio and aistor to point to same structs, should this still be moved to `madmin-go` ? @harshavardhana

PR #137, opened 2024-10-25
Always use offline public key for license verification

ramondeklein:
The MinIO console attempts to verify the license key. It needs the public key to verify the license signature and it downloads the key from subnet. MinIO is not supposed to contact external sites, unless explicitly asked to do so (https://github.com/minio/console/issues/3456).

It now always uses the embedded public key (like it already did when subnet couldn't be reached). All license based products will be moving to AIStor eventually, so having the ability to deal with a rotating public key is less important.

### DIFF
--- a/subnet/license.go
+++ b/subnet/license.go
@@ -18,7 +18,6 @@
 package subnet
 
 import (
-	"bytes"
 	"context"
 	"crypto/tls"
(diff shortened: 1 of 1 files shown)


anjalshireesh (approved):
LGTM

PR #138, opened 2024-11-13
Add policy name to audit log entries

marktheunissen:
Required by: https://github.com/minio/minio/pull/20638


marktheunissen:
Closing as we will use tags instead

PR #146, opened 2024-12-19
Add the SiteName field to the audit struct

allanrogerr:
This is in preparation for the addition of the SiteName to both internal and external audit logs


klauspost:
@allanrogerr Please state why. It seems silly to add a property that will be the same for every audit log entry.  You already have the deployment ID as a site separator.

harshavardhana:
this package is being moved to `madmin-go`

PR #152, opened 2025-01-09
Add nil check for certificate listing

shtripat:


vadmeste (approved):

certs/manager.go
```diff
@@ -447,7 +447,7 @@ func (m *Manager) GetAllCertificates() []*x509.Certificate {
 
 	certs := []*x509.Certificate{}
 	for _, c := range m.certificates {
-		if c.Leaf != nil {
+		if c != nil && c.Leaf != nil {
```
vadmeste, line 450: does this fix anything ? it does not seem it is possible to have this nil in the code

shtripat: This may not be needed as Harsha already fixed in previous PR.

PR #155, opened 2025-03-20
Allow disabling auto-reload

ramondeklein:
The current `certs.Manager` only allows reloading of existing certificates. These certificates are reloaded when the certificate file (on disk) is changed or when the manager receives `SIGHUP`.

For AIStor operator, we should also support adding new certificates. The sidecar will copy the certificates to the appropriate location and instruct MinIO to reload **ALL** certificates by sending `SIGHUP`. By disabling the auto-reload functionality in `certs.Manager`, reloading the certificates twice (and possible dangling Go-routines) will be prevented.

If `certs.WithDisableAutoReload()` is not added, then functionality is still the same.

This is required for https://github.com/miniohq/eos/pull/554.

### DIFF
--- a/certs/manager.go
+++ b/certs/manager.go
@@ -43,10 +43,11 @@ import (
 //
 // Manager will automatically reload certificates if the corresponding file changes.
 type Manager struct {
-	lock         sync.RWMutex
-	certificates map[pair]*tls.Certificate // Mapping: certificate file name => TLS certificates
-	defaultCert  pair
-	duration     time.Duration
+	lock              sync.RWMutex
+	certificates      map[pair]*tls.Certificate // Mapping: certificate file name => TLS certificates
+	defaultCert       pair
+	duration          time.Duration
+	disableAutoReload bool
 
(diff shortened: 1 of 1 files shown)


ramondeklein:
@aead @harshavardhana I would like to go ahead with my `SIGHUP` reload PR. It's done, but this PR needs to be merged first. I think it's fully backwards compatible, but I would like to have your blessing and get this merged. Then we can review/approve https://github.com/miniohq/eos/pull/554.

ramondeklein:
@aead @harshavardhana Can you review? We need this fix to allow the operator sidecars to reload all certificates when the `ObjectStore` configuration changes.

ramondeklein:
@harshavardhana Do you know anyone else who can review this PR? I would like to improve the auto-reloading of certificates in Kubernetes without restarting. I had to restart a customer's statefulset today again, because they added an external certificate and it wasn't picked up.

harshavardhana (approved):

PR #157, opened 2025-04-11
chore: Get Next Immediate Port

sivanantham-gnanavel:
Context: https://github.com/miniohq/eos/issues/646 - This change introduces a reusable function for use in https://github.com/miniohq/eos/pull/649.

### DIFF
--- a/net/port.go
+++ b/net/port.go
@@ -46,6 +46,25 @@ func GetFreePort() (Port, error) {
 	return Port(l.Addr().(*net.TCPAddr).Port), nil
 }
 
+// GetNextPort returns the immediate next port if it's available.
+func GetNextPort(port string) (Port, error) {
+	if port == "" || port == "0" {
+		return 0, errors.New("invalid starting port")
+	}
+	p, err := strconv.Atoi(port)
+	if err != nil || p <= 0 {
+		return 0, errors.New("invalid port number")
+	}
+	nextPort := p + 1
+	addr := &net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: nextPort}
+	l, err := net.ListenTCP("tcp", addr)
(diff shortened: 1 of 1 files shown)


net/port.go
```diff
@@ -46,6 +46,25 @@ func GetFreePort() (Port, error) {
 	return Port(l.Addr().(*net.TCPAddr).Port), nil
 }
 
+// GetNextPort returns the immediate next port if it's available.
+func GetNextPort(port string) (Port, error) {
```
harshavardhana, line 50:
```suggestion
func NextFreePort(port string) (Port, error) {
```

usage wise its correct 

```
xnet.NextFreePort()
```

sivanantham-gnanavel: We have another function name called `GetFreePort()`. To follow the same can I named as `GetNextFreePort()`

net/port.go
```diff
@@ -46,6 +46,25 @@ func GetFreePort() (Port, error) {
 	return Port(l.Addr().(*net.TCPAddr).Port), nil
 }
 
+// GetNextFreePort returns the immediate next port if it's available.
+func GetNextFreePort(port string) (Port, error) {
+	if port == "" || port == "0" {
+		return 0, errors.New("invalid starting port")
+	}
+	p, err := strconv.Atoi(port)
+	if err != nil || p <= 0 {
```
jkandasa, line 55: validate the upper range too

sivanantham-gnanavel: @jkandasa I've added the upper limit validation

jkandasa:
LGTM

net/port.go
harshavardhana: yeah

harshavardhana (approved):

PR #160, opened 2025-05-01
adding admin:NodeList, NodeInfo and DriveList

zveinn:
Adding NodeList, NodeInfo and DriveList

DriveList is currently not in use inside minio, but will be implemented with the new v4 apis


harshavardhana (changes requested):
These APIs must be added SupportedAdminAction list as well

harshavardhana (approved):

PR #173, opened 2025-06-26
fix: incorrect policy evaluation

donatello:
Regression from #167


policy/policy.go
```diff
@@ -208,7 +208,6 @@ func (iamp Policy) IsAllowed(args Args) bool {
 				}
 			}
 		}
-		return false
```
donatello, line 211: This is the fix.

harshavardhana (approved):

taran-p (approved):

PR #185, opened 2025-08-21
Add inventory API privileges

donatello:


policy/action.go
```diff
@@ -206,6 +206,13 @@ const (
 	// PutObjectFanOutAction - PutObject like API action but allows PostUpload() fan-out.
 	PutObjectFanOutAction = "s3:PutObjectFanOut"
 
+	// Inventory configuration actions
+
+	// PutInventoryConfigurationAction - Bucket inventory write operations actions
+	PutInventoryConfigurationAction = "s3:PutInventoryConfiguration"
+	// GetInventoryConfigurationAction - Bucket inventory read operations actions
+	GetInventoryConfigurationAction = "s3:GetInventoryConfiguration"
```
harshavardhana, line 214: You don't need this for ListInventory jobs? DeleteInventory?

harshavardhana (changes requested):

policy/action.go
donatello: It's kept simple and similar to AWS S3 here: Put action is used for all mutations (like PutBucketInventoryConfiguration and DeleteBucketInventoryConfiguration) and Get action for all non-mutations (like GetBucketInventoryConfiguration, GenerateBucketInventoryConfiguration, GetBucketInventoryJobStatus and ListBucketInventoryConfigurations)

PR #186, opened 2025-08-21
Deny CreateUser actions in built in policies

reivaj05:
Part of: https://github.com/miniohq/aistor-console/issues/4705


harshavardhana:
I don't think we can change existing policies anymore, canned policies cannot be extended anymore. This is a feature requirement if they want to disable creating users customers might explicitly add a Deny rule.

reivaj05:
That's ok but we still need a way to know if the action is not allowed, right now the code is returning that the action CreateUser is allowed for all users, this is confusing in the ui since a user without policies will have this action allowed by default but that user can't really create users

PR #188, opened 2025-09-03
Cordon action permission added

marktheunissen:


klauspost (approved):

harshavardhana (changes requested):
Not just here you need to update the `supportedAdminActions` as well

harshavardhana:
> `HealAdminAction` and all the rest of these consts are type `string`, not type `AdminAction`. I can change these if wanted. They are used interchangeably (AdminAction and Action), in a few places. So I don't think compiler type checking is actually enforcing anything, but it is a bit misleading, especially since the first `Action`, `AbortMultipartUploadAction`, IS actually of type `Action` but the rest are not.

Correct all these must be fixed @marktheunissen

marktheunissen:
@harshavardhana what about adding GenerateBatchJobAction and ForceUnlockAdminAction?

PR #190, opened 2025-09-15
Add lifecycle common code to mc and console

reivaj05:
Adds code related to lifecycle management that can be re used in console and mc


harshavardhana (changes requested):
The changes look good, but please add unit tests - we can't merge without it.

reivaj05:
@harshavardhana unit tests added, ptal

harshavardhana (approved):

PR #193, opened 2025-09-30
Expose policy constants for actions, conditions

reivaj05:
We are working a policy builder in console, for this we need to expose these constants so they can use to populate the different inputs in the UI component

Part of https://github.com/miniohq/aistor-console/issues/4911

Console implementation: https://github.com/miniohq/aistor-console/pull/4912


reivaj05:
@harshavardhana Could you please take a look at this PR? 👀

PR #232, opened 2026-04-10
tables: add a policy action for catalog scanner debugging actions like reseting the replica

marktheunissen:


krisis:
@marktheunissen could we include the various IAM permissions and the corresponding tables replication action they authorize? The replication design draft wiki would be a good place to have this documented in.

krisis (approved):

marktheunissen:
These two were already there added by Shubhendu, I think they could be removed as Shubhendu confirmed that they are not used.

```
	TablesReplicationAddAction:           {},
	TablesReplicationRemoveAction:        {},
```
yes the remaining ones I will document

twuebi (approved):

marktheunissen:
@harshavardhana PTAL

klauspost (approved):

PR #249, opened 2026-08-01
feat(policy): add the arn:minio:memory ARN for Memory resources

harshavardhana:
## What

Memory resources get their own ARN namespace:

```
arn:minio:memory:::<cortex>[/<collection>[/<name>]]
```

They were authorized against the S3 ARN of the bucket backing the cortex, which
forced the storage path into every policy and made a Memory grant
indistinguishable from raw S3 access to the same bytes. The pattern is now a
logical path, so the Memory API's on-disk layout can change without
invalidating a policy.

`ValidateMemory` requires the Memory type, so an S3 ARN cannot carry a Memory
action and a Memory ARN cannot carry an S3 one.

## Bare ARN prefix fix (affects every ARN type)

`arn:aws:s3:::` parsed as `ResourceARNAll` keeping its own prefix as the
pattern. It validated cleanly and then matched nothing — an `Allow` that
granted zero and, the direction that fails open, a `Deny` that never fired:

```
ParseResource("arn:aws:s3:::") -> Type=* Pattern="arn:aws:s3:::"
    matches "mybucket/key"? false
policy validates: <nil>
GetObject allowed despite the Deny? true
```

AWS rejects the form outright. It now parses as its own type with an empty
pattern. S3, S3 Tables and KMS keep loading one so a policy already on disk
still works, with `ValidateStrict` refusing it on create/update — the split
`ParseConfigStrict` already exists to provide. Memory carries no such
tolerance and fails at parse.

## Tests

Memory resources had no coverage; the suite passed because nothing exercised
them. Adds parsing, validation, matching, cross-type isolation, whole-document
validation and end-to-end evaluation — including that an agent-scoped grant
denies a neighbour whose id shares a prefix, and that a bare-prefix policy
loads but cannot be written.

Both defences mutation-checked: dropping traversal/charset rejection fails 17
assertions; letting Memory inherit the bare-ARN tolerance fails 3.

## Summary by CodeRabbit

- **New Features**
  - Added support for AIStor Memory resources in policy ARNs.
  - Added Memory resource validation and wildcard matching.
  - Added strict policy validation that rejects incomplete resource prefixes and accepts explicit wildcard forms.
  - Added safeguards for resource type and scope isolation.

- **Bug Fixes**
  - Prevented malformed or traversal-based Memory resources from being accepted.
  - Ensured resource representations do not expose underlying storage details.
  - Preserved compatibility when loading existing bare ARN policies while preventing unsafe updates.

### DIFF
--- a/policy/memory-resource_test.go
+++ b/policy/memory-resource_test.go
@@ -157,7 +157,7 @@ func TestBareARNPrefixIsNotWildcard(t *testing.T) {
 			t.Errorf("ParseResource(%q): pattern = %q, want empty", prefix, r.Pattern)
 		}
 
-		// Inert: it matches nothing, which is the behaviour already on disk.
+		// Inert: it matches nothing, which is the behavior already on disk.
 		for _, probe := range []string{"bucket", "bucket/key", prefix, "*"} {
 			if r.MatchResource(probe) {
 				t.Errorf("bare %q matched %q; it must match nothing", prefix, probe)
@@ -246,7 +246,7 @@ func TestBareARNPolicyLoadsButCannotBeWritten(t *testing.T) {
 		BucketName:  "secret",
 		ObjectName:  "data",
 	}) {
-		t.Error("behaviour of an already-stored policy changed")
+		t.Error("behavior of an already-stored policy changed")
 	}
 }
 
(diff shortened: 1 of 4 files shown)


harshavardhana:
Pushed `98d63e9`. All four findings actioned.

**Bare prefix breaks cross-type loading** (outside-diff, Major) — correct, and it defeated the point of the tolerance. Reproduced before fixing:

```
KMS statement with bare S3 prefix, Validate() = resource 'arn:aws:s3:::' type is not KMS
```

`ValidateS3`/`KMS`/`Table`/`Vectors` now tolerate an inert bare resource. `ValidateMemory` deliberately does not — a Memory statement carrying one cannot already exist on disk, so there is nothing to keep loading. `TestBareARNLoadsInsideForeignStatement` covers all four combinations.

**Traversal not enforced** (Major) — agreed, and it is enforceable for Memory alone without touching S3 semantics. `Match` now cleans the candidate when the type is Memory. The test is a positive assertion instead of a self-skip, and also checks that cleaning does not break resources legitimately inside the subtree, and that a path resolving *into* the subtree still matches.

S3, Tables and KMS still match verbatim — that is pre-existing and changing it has a much wider blast radius, so it is documented on `Match` rather than altered here.

**misspell** — fixed all three. Also ran `make lint` locally this time; 0 issues.

**Subtests** — both policy tables now use `t.Run`, plus a valid Memory `NotResource` case.

harshavardhana:
**Correction.** This PR's description and commit message state that AWS rejects `arn:aws:s3:::` outright. That is unsupported and I should not have asserted it. #250 corrects the code comment carrying the same claim.

Checked offline against three validators:

- **moto** — reimplements AWS's `MalformedPolicyDocument` responses. It **accepts** `arn:aws:s3:::`. It rejects `arn:aws:s3` ("can not contain region information") and `notanarn` ("must be in ARN format or `*`"), so it does have a resource-ARN rejection path; the empty form passes through it.
- **parliament** — its `INVALID_ARN` check tests part count, prefix, partition, region and account shape. It never checks the resource segment is non-empty.
- **policy_sentry** — parses it cleanly as `{service:s3, region:'', account:'', resource:''}`.

AWS's [policy grammar](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html) defines `resource_string` as only "In most cases, consists of an ARN", so spec-malformedness cannot be argued from it either.

What does hold, and what the corrected comment says: AWS documents no S3 resource type with an empty bucket name, and its wildcard-completion rule is scoped to ARNs with *fewer than six fields* — `arn:aws:s3:::` has six, so it is not equivalent to `arn:aws:s3:::*`.

**The fix is unaffected.** It rests on MinIO's own behavior, reproduced in the description above: a bare prefix parsed as `ResourceARNAll` keeping its own prefix as the pattern, matched nothing, and a `Deny` written that way never fired. That repro stands on its own and never depended on AWS rejecting anything.

PR #256, opened 2026-08-09
feat(policy): add MemoryGetObjectBioAction

harshavardhana:
## Description

Adds `memory:GetObjectBio` to the Memory API action set.

## Motivation

AIStor's Memory API is growing an object-bio route that returns one object's
version history and per-version provenance. Every other Memory surface
authorizes a `memory:*` action via `authorizeMemoryAction`; without this
constant the route has to borrow an S3 action, which grants the caller more
than the route needs and makes it the one inconsistent member of the family.

Registered as a read rather than an enumeration: it names one record in its
resource and has no query to constrain, so `MemoryPrefix`/`MemoryMaxKeys` do
not apply.

## How to test

`go test ./policy/ -run Memory` — `TestMemoryActionIsValid` covers the new
constant.

## Summary by CodeRabbit

- **New Features**
  - Added support for the `memory:GetObjectBio` action.
  - The action is now recognized as valid and available for action-based conditions.

### DIFF
--- a/policy/memory-action.go
+++ b/policy/memory-action.go
@@ -65,6 +65,12 @@ const (
 	// MemorySearchAction - search (corpus-grep) the objects in a cortex.
 	MemorySearchAction MemoryAction = "memory:Search"
 
+	// MemoryGetObjectBioAction - read an object's biography from a cortex:
+	// who authored each version of it, when, from which agent and session,
+	// and what that version carries. Distinct from reading the object, which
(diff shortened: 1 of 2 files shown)


harshavardhana:
Closing in favour of #255, which Dil opened first and which carries the same change. I pushed the doc-comment wording onto that branch instead — an object biography is who authored each version, when, from which agent and session, not a version history.

PR #260, opened 2026-08-20
feat(policy): let memory:prefix scope memory:Search

harshavardhana:
`memory:prefix` already exists and already scopes the list actions. `memory:Search` was left out, and it is the action that needs it most: a list returns names, a search returns the **content** of every object under its prefix. So the grant has exactly two states today — no search at all, or search over every object in the cortex.

That matters wherever one cortex holds more than one tenant's data. A credential scoped to a single prefix can read only its own objects over S3, and then either cannot search at all or can grep everyone's.

## Measured, not inferred

Against a live server, minting STS sessions with inline policies over a cortex holding `workspaces/alice/app-7/` and `workspaces/bob/app-9/`:

| session policy | result |
| --- | --- |
| cortex ARN + `arn:minio:memory:::<cortex>/workspaces/alice/app-7/*` | search runs, and a request for **bob's** prefix returns bob's content |
| `arn:minio:memory:::<cortex>/workspaces/alice/app-7/*` alone | 403 on every search, including the session's own prefix |
| cortex ARN alone | search runs over the whole cortex |

The key-level resource is inert for this action: present it grants nothing, absent nothing is lost.

## Why the condition key rather than the resource

`condition.MemoryPrefix`'s own doc comment already makes the argument for lists, and it transfers unchanged — a search names a query, not a record, so putting the prefix in the resource path would make one ARN mean a single object under `memory:GetObjectBio` and every object sharing that prefix under `memory:Search`. This is the same relationship `s3:prefix` has with `s3:ListBucket`.

## The change

`MemorySearchAction` joins the actions that accept the enumeration condition keys, so a policy can say:

```json
{
  "Effect": "Allow",
  "Action": ["memory:Search"],
  "Resource": ["arn:minio:memory:::orb"],
  "Condition": {"StringLike": {"memory:prefix": ["workspaces/alice/app-7/*"]}}
}
```

Additive. A statement carrying no condition behaves exactly as it does today, and the point actions still reject the key — `memory:GetObjectBio` is now covered by that half of `TestMemoryEnumerationConditionKeys`, having been absent from it.

## Server side

The server must populate the value and, critically, search the same prefix it authorized. That lands separately in aistor: the search prefix moves into the query string (the only place `getConditionValues` looks), and a request whose query and body prefixes disagree is refused rather than reconciled — authorizing one prefix and walking a wider one is the failure worth guarding against.

Verified end to end against a server carrying both changes: a session scoped to one workspace searches its own prefix, is refused another's, and is refused a cortex-wide search; an unconditioned session behaves exactly as before.

`go test ./policy/...` passes.

## Summary by CodeRabbit

- **New Features**
  - Added support for scoping memory searches with memory-prefix conditions.
  - Added maximum-key conditions to memory search actions.
  - Improved policy controls for workspace-scoped memory search and unrestricted content access.

- **Documentation**
  - Clarified how memory-prefix conditions apply to memory listing and search APIs.

### DIFF
--- a/policy/condition/keyname.go
+++ b/policy/condition/keyname.go
@@ -134,10 +134,14 @@ const (
 	S3Delimiter KeyName = "s3:delimiter"
 
 	// MemoryPrefix - key representing the prefix query parameter of the AIStor
-	// Memory list APIs only. It scopes enumeration within a cortex, which a
-	// resource ARN cannot express: a list names a query, not a record, so
+	// Memory list and search APIs. It scopes enumeration within a cortex, which
+	// a resource ARN cannot express: a list names a query, not a record, so
 	// putting the prefix in the resource path would make one ARN mean a single
 	// record under a read action and a set of siblings under a list action.
+	//
+	// memory:Search takes it for the same reason and depends on it more, since
+	// search returns object content: without the key its grant has only two
+	// states, none or the whole cortex.
 	MemoryPrefix KeyName = "memory:prefix"
 
 	// MemoryMaxKeys - key representing the limit query parameter of the AIStor
--- a/policy/memory-action.go
+++ b/policy/memory-action.go
@@ -109,14 +109,22 @@ func createMemoryActionConditionKeyMap() map[Action]condition.KeySet {
 		memoryActionConditionKeyMap[Action(act)] = condition.NewKeySet(commonKeys...)
 	}
 
-	// Enumeration keys apply only to the list actions. A read or a write names
-	// one record in its resource and has no query to constrain, so accepting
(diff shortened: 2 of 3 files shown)


harshavardhana:
Addressed in e2e1910: the rejection half of `TestMemoryEnumerationConditionKeys` now iterates both `memory:prefix` and `memory:max-keys` over every point action, matching the acceptance half. It had checked only the prefix key, so nothing showed that max-keys hadn't been quietly admitted on a read or a write.

`go test ./policy/...` passes.

harshavardhana:
@coderabbitai review

PR #261, opened 2026-08-21
Fix policy validation & improve speed

klauspost:
PR 1 — security

## fix(policy,wildcard): bound wildcard matching and close two Deny-dropping paths

Three independent correctness bugs on the authorization path. No API changes.

### 1. `wildcard.Match` was exponential in the number of `*`

`deepMatchRune` recursed on both alternatives at every `*`, so its cost grew exponentially with the star count. The pattern is caller-supplied — statement validation classifies every action against every namespace, and `Resource.Match` runs untrusted patterns — so a tiny document could consume unbounded CPU:

| input                                                     | before  | after  |
| --------------------------------------------------------- | ------- | ------ |
| `ParseConfig`, action `"*********x"`                      | 59.3 s  | < 1 ms |
| `ParseBucketPolicyConfig`, same action                     | 18.8 s  | < 1 ms |
| `AdminAction.IsValid`, 8 stars                            | 4.5 s   | < 1 ms |

Scaling was ~4x per added star: 6 stars 165 ms, 8 stars 4.5 s, 10 stars > 100 s. `MatchSimple` was equally affected, and it backs bucket-policy principal matching and CORS origin matching.

Replaced with the standard greedy two-pointer matcher: one backtrack point, no recursion, no allocation, O(n*m) worst case.

`MatchSimple`'s documented "a trailing `?` is optional" behaviour does not survive greedy backtracking — a first attempt that folded it into the matcher diverged on `MatchSimple("*?*a", "a")`. It is now expressed explicitly in `MatchSimple`: if the pattern does not match outright, any prefix of it ending at a `?` that consumes the whole name also matches. The matcher itself is now plain glob semantics with no mode flag.

### 2. `Statement.hash` omitted `NotResources`

`dropDuplicateStatements` switches to hashing above 10 statements. Two statements differing *only* in `NotResource` hashed identically, so one was discarded at parse time. When those are `Deny` statements, a restriction silently disappears from a policy that was written correctly.

Reproduced: an 11-statement policy carrying two distinct `Deny ... NotResource` statements loses one; the same policy with 4 statements (below the threshold, exact-comparison path) keeps both.

### 3. `HasDenyStatement` trusted a field that struct-literal policies never set

`hasDeny` is only assigned in `updateActionIndex`, which runs from `UnmarshalJSON` and `MergePolicies`. A `Policy` built as a struct literal never reaches it — including this package's own `readonly` and `consolereadonly` defaults in `constants.go`, both of which carry `Deny admin:CreateUser`. `HasDenyStatement()` reported `false` for them.

That matters to callers that use the answer to route policies: AIStor sorts them into a "no deny statements" bucket, evaluates that bucket with an early return on the first `Allow`, and never reaches the `Deny`. A principal holding `readonly` plus any policy allowing `admin:CreateUser` was granted `admin:CreateUser`.

`HasDenyStatement` now falls back to scanning statements. It runs once per policy per merge, not per request.

### Compatibility

Fix 2 changes which statements survive parsing. A policy that was accidentally relying on the collapse regains a `Deny`, so a request that previously succeeded may start returning 403. That is the intended direction, but it belongs in release notes.

Fixes 1 and 3 are behaviour-preserving for any correctly-built input.

### Testing

- `wildcard/match_equivalence_test.go` keeps the previous recursive matcher as a reference implementation and diffs against it: 254,512 exhaustive pattern/name/mode combinations (patterns over `{a,b,*,?}` to length 5, `{a,:,/}` names) plus a differential fuzz target. 101M fuzz executions, no divergence.
- `TestMatchStarsAreLinear` fails if a single `Match` with 8/16/64/256 stars, or 32 interleaved `*a` pairs, exceeds 50 ms.
- `policy/parse_dos_test.go` asserts `ParseConfig` and `ParseBucketPolicyConfig` stay under 2 s for star-heavy actions.
- `TestDropDuplicateStatementsKeepsNotResources` covers both the exact-comparison and hashed paths.
- `TestHasDenyStatementOnStructLiteralPolicy` walks the built-in defaults.

`go test ./...` is green.

PR 2 — performance

## perf(policy): cut authorization cost for large policy collections

A production profile of a 24-node cluster found 61–73% of all CPU on client-facing nodes inside `Policy.Decide`, and 51–61% in `AdminAction.IsValid` alone. The credential resolved to more than 4000 policies. Six changes, no API changes, no behaviour changes.

### `AdminAction.IsValid` / `HasResource` — O(1) fast path

Both scanned their whole action map calling `wildcard.Match` per entry, with the receiver as the pattern. `Statement.isAdmin` calls `IsValid` for every action of every statement it evaluates, so a pure-S3 statement paid `len(Actions) x 101` wildcard matches to learn nothing.

Now: exact map lookup; then bail if the pattern holds no metacharacter (a literal can only match by being in the map); then bail if the literal head before the first metacharacter is prefix-incompatible with `admin:`.

| | before | after |
| ---------------------- | ------- | ------- |
| `isAdmin`, 1 action    | 690 ns  | 34.9 ns |
| `isAdmin`, 3 actions   | 2012 ns | 63.9 ns |

The head comparison is bidirectional on purpose: `adm*` and `a*d*m*i*n*` are both valid admin patterns, so a plain `strings.HasPrefix(head, "admin:")` would be a silent authorization change.

### `actionStatementIndex` was a pessimisation

`Decide` consulted the index and then — hit or miss — fell through to an unconditional walk of every statement, re-evaluating everything the index had just evaluated. The index only ever helped when an indexed statement allowed. The walk now skips the positions the index already tried (`indexes` is ascending, and `isAllowedFor` is pure).

### One classification pass instead of five, cached at parse time

`IsAllowedPtr` called `isTable`, `isKMS`, `isSTS`, `isAdmin` and `hasAdminResource`, each ranging over the action set separately — four map-iterator setups per statement per authorization. Collapsed into one `classify()` pass, and cached on the statement by `updateActionIndex`.

The cached field's zero value means "not computed — derive it now", not "no namespace". A statement built as a struct literal, or cloned, is therefore slower rather than wrong, and `Statement.Clone` deliberately does not copy it. `classify` is a pure function of `Statement.Actions`, which nothing mutates after parse — the same invariant `actionStatementIndex` already relies on.

### The request resource string is built once per request

`IsAllowedPtr` rendered `BucketName`/`ObjectName` into a pooled buffer and then called `Buffer.String()` — which copies — once or twice *per statement*. A memory profile of one authorization over 156 policies of 20 statements put 92.5% of all allocations on that one line: 1561 allocations and 75 KB of garbage per authorization.

It depends only on the request, so it is now memoized on `Args` and passed to every statement. `Args` carries the fields it was built from, so a caller reusing an `Args` for a second object cannot read a stale value.

Note: one `*Args` must not be shared across goroutines within a single evaluation. `IsAllowedPar` did share one; it now gives each worker its own copy.

### Measured

`go test ./policy/ -run XXX -bench 'BenchmarkIsAllowed|BenchmarkSerialEvalVsParEval' -benchmem`, on an idle 32-thread box:

| benchmark                            | before   | after    |       |
| ------------------------------------ | -------- | -------- | ----- |
| `IsAllowed/SingleStatementAllow`     | 1811 ns  | 127 ns   | 14.3x |
| `IsAllowed/DenyRule`                 | 2737 ns  | 224 ns   | 12.2x |
| `IsAllowed/WildcardMatching`         | 919 ns   | 121 ns   | 7.6x  |
| `IsAllowed/MultipleStatements`       | 6054 ns  | 4069 ns  | 1.5x  |
| `SerialEvalVsParEval/128p` serial    | 1.62 ms  | 252 us   | 6.4x  |
| `SerialEvalVsParEval/1024p` serial   | 12.6 ms  | 2.07 ms  | 6.1x  |

`MultipleStatements` gains least because its statements have non-matching action sets, so `Actions.Match` short-circuits before classification runs. `SerialEvalVsParEval` builds policies as struct literals and never calls `updateActionIndex`, so it does not exercise the cached classification and understates the result — measured through `ParseConfig`, as a server loads policies, the same shape at 4000 policies goes from 66.1 ms to 660 us per authorization with allocations dropping from 88,888 to 1.

Allocation caveat: benchmarks with short bucket+key strings previously stack-allocated the per-statement copy and report 0 -> 1 allocs/op. Anything with a realistic object key allocated once per statement before and once per request now.

### Testing

- `policy/admin-action_test.go` (new): trap table for the fast path (`"*"` -> true, `"a*d*m*i*n*"` -> true, `"admin:heal"` ->
  false, `"s3:*"` -> false, ...), the same for `HasResource`, an equivalence check against the previous linear scan over 964
  patterns, and a differential fuzz target.
- `TestAdminActionNamespacePrefix` pins the invariant the fast path rests on — every `SupportedAdminActions` key starts with `admin:`. Nothing enforced this before, and the analogous `s3:` invariant is *already* broken by `s3express:CreateSession` in `SupportedActions`, so it can rot into an authorization change.
- `TestDecideReachesDenyOnlyAndIsOwnerWithNoStatements`: `Decide`'s `DenyOnly` and `IsOwner` returns sit below the deny loop, so a policy with nothing to match must still reach them. An earlier draft of this change added an early return for the empty case and silently turned both from allow into deny. Any future work narrowing which statements `Decide` walks needs this test.

## Summary by CodeRabbit

- **Performance**
  - Improved policy authorization speed through more efficient resource and action evaluation.
  - Accelerated wildcard matching, including complex patterns with many wildcards.
  - Reduced delays when processing complex policy action patterns.

- **Bug Fixes**
  - Improved handling of deny statements, empty policies, and policy updates.
  - Preserved distinct policy statements with different excluded resources.
  - Improved consistency when policies are re-evaluated.

- **Tests**
  - Added coverage for authorization decisions, action validation, wildcard matching, and performance safeguards.

### DIFF
--- a/policy/admin-action.go
+++ b/policy/admin-action.go
@@ -18,6 +18,8 @@
 package policy
 
 import (
+	"strings"
+
 	"github.com/minio/pkg/v3/policy/condition"
 	"github.com/minio/pkg/v3/wildcard"
 )
@@ -451,6 +453,13 @@ var AdminActionsWithResource = map[AdminAction]struct{}{
 
 // HasResource reports whether this admin action operates on a bucket resource.
 func (action AdminAction) HasResource() bool {
+	if _, ok := AdminActionsWithResource[action]; ok {
+		return true
+	}
+	if !wildcard.Has(string(action)) {
+		// A literal action can only match by being in the set.
+		return false
+	}
 	for a := range AdminActionsWithResource {
 		if action.Match(a) {
 			return true
@@ -465,13 +474,45 @@ func (action AdminAction) Match(a AdminAction) bool {
 }
 
 // IsValid - checks if action is valid or not.
+//
+// The receiver is the pattern, so this asks whether the pattern matches any
+// supported admin action. Statement.isAdmin calls it for every action of every
+// statement it evaluates, so the two cases that can be answered without
+// touching SupportedAdminActions are answered first: a literal action can only
+// match by being in the set, and a pattern is only worth scanning when its
+// literal head is prefix-compatible with the admin namespace.
 func (action AdminAction) IsValid() bool {
+	if _, ok := SupportedAdminActions[action]; ok {
+		return true
+	}
+	s := string(action)
+	star := strings.IndexAny(s, "*?")
+	if star < 0 {
+		return false
+	}
+	if !canMatchPrefix(s[:star], adminActionPrefix) {
+		return false
+	}
 	for supAction := range SupportedAdminActions {
 		if action.Match(supAction) {
 			return true
 		}
 	}
 	return false
+}
+
+// adminActionPrefix is the namespace every supported admin action carries.
+const adminActionPrefix = "admin:"
+
+// canMatchPrefix reports whether a wildcard pattern whose literal head (the
+// part before its first metacharacter) is head could match any string starting
+// with prefix. Everything up to the first metacharacter has to match literally,
+// so head and prefix must agree wherever they overlap.
+func canMatchPrefix(head, prefix string) bool {
+	if len(head) > len(prefix) {
+		return strings.HasPrefix(head, prefix)
+	}
+	return strings.HasPrefix(prefix, head)
 }
 
 func createAdminActionConditionKeyMap() map[Action]condition.KeySet {
--- a/policy/admin-action.go
+++ b/policy/admin-action.go
@@ -476,11 +476,7 @@ func (action AdminAction) Match(a AdminAction) bool {
 // IsValid - checks if action is valid or not.
 //
 // The receiver is the pattern, so this asks whether the pattern matches any
-// supported admin action. Statement.isAdmin calls it for every action of every
-// statement it evaluates, so the two cases that can be answered without
-// touching SupportedAdminActions are answered first: a literal action can only
-// match by being in the set, and a pattern is only worth scanning when its
-// literal head is prefix-compatible with the admin namespace.
+// supported admin action.
 func (action AdminAction) IsValid() bool {
 	if _, ok := SupportedAdminActions[action]; ok {
 		return true
@@ -488,9 +484,13 @@ func (action AdminAction) IsValid() bool {
 	s := string(action)
 	star := strings.IndexAny(s, "*?")
 	if star < 0 {
+		// A literal action can only match by being in the set above.
 		return false
 	}
 	if !canMatchPrefix(s[:star], adminActionPrefix) {
+		// computeClass runs this for every action of every statement it
+		// classifies, so a pattern is only scanned against the whole set when
+		// its literal head is prefix-compatible with the admin namespace.
 		return false
 	}
 	for supAction := range SupportedAdminActions {
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -569,12 +569,9 @@ func (iamp *Policy) Reindex() {
 	iamp.updateActionIndex()
 }
 
-// updateActionIndex with latest statements()
-// maintains a reverse map of Action -> []Statements
-// for faster lookup and short-circuit.
-//
-// Everything it derives is discarded first, so calling it a second time
-// replaces the previous result rather than accumulating onto it.
+// updateActionIndex fills the Action -> []Statements reverse lookup by
+// appending, so it discards the derived state first: a repeat call has to
+// replace the previous result rather than accumulate onto it.
 func (iamp *Policy) updateActionIndex() {
 	iamp.actionStatementIndex = nil
 	iamp.hasDeny = false
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -46,6 +46,24 @@ type Args struct {
 	ObjectName      string              `json:"object"`
 	Claims          map[string]any      `json:"claims"`
 	DenyOnly        bool                `json:"denyOnly"` // only applies deny
+
+	// Memoized requestResource output, with the fields it was built from so a
+	// caller that reuses an Args for a second object cannot read a stale one.
+	// A single evaluation must not share one *Args across goroutines.
+	resource     string
+	resourceFrom [2]string
+}
+
+// requestResource returns the resource string this request names, building it
+// at most once per Args. Every statement of every policy needs the same string,
+// and rebuilding it per statement was the largest single source of allocation
+// on the authorization path.
+func (a *Args) requestResource() string {
+	if a.resource == "" || a.resourceFrom[0] != a.BucketName || a.resourceFrom[1] != a.ObjectName {
+		a.resource = buildRequestResource(a)
+		a.resourceFrom = [2]string{a.BucketName, a.ObjectName}
+	}
+	return a.resource
 }
 
 // GetValuesFromClaims returns the list of values for the input claimName.
@@ -126,8 +144,21 @@ type Policy struct {
 }
 
 // HasDenyStatement returns if the policy has a deny statement.
+//
+// hasDeny is only populated by updateActionIndex, which a policy built as a
+// struct literal never reaches — the built-in readonly policies are built that
+// way and do carry a Deny — so the statements are the authority and the field
+// is only a shortcut.
 func (iamp *Policy) HasDenyStatement() bool {
-	return iamp.hasDeny
+	if iamp.hasDeny {
+		return true
+	}
+	for i := range iamp.Statements {
+		if iamp.Statements[i].Effect == Deny {
+			return true
+		}
+	}
+	return false
 }
 
 // MatchResource matches resource with match resource patterns
@@ -247,8 +278,10 @@ func IsAllowedPar(policies []Policy, args Args) bool {
 
 				maxJ := min(i+numPoliciesPerWorker, len(policies))
 				res := NoDecision
+				// Args memoizes per-request state, so each worker needs its own.
+				a := args
 				for j := i; j < maxJ; j++ {
-					decision := policies[j].Decide(&args)
+					decision := policies[j].Decide(&a)
 					if decision == DenyDecision {
 						res = DenyDecision
 						break
@@ -298,9 +331,12 @@ const (
 // statement explicitly allows or denies the operation in the Args, it returns
 // `noDecision`. It is upto the caller to handle such cases.
 func (iamp *Policy) Decide(args *Args) Decision {
+	resource := args.requestResource()
+
 	// Check all deny statements. If any one statement denies, return false.
-	for _, statement := range iamp.Statements {
-		if statement.Effect == Deny && !statement.IsAllowedPtr(args) {
+	for i := range iamp.Statements {
+		statement := &iamp.Statements[i]
+		if statement.Effect == Deny && !statement.isAllowedFor(args, resource) {
 			return DenyDecision
 		}
 	}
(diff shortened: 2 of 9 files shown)


policy/policy.go
```diff
@@ -126,8 +144,21 @@ type Policy struct {
...
-	return iamp.hasDeny
+	if iamp.hasDeny {
+		return true
+	}
+	for i := range iamp.Statements {
+		if iamp.Statements[i].Effect == Deny {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
klauspost: Policy is exported and callers assemble literals directly; trusting
  the flag alone would silently under-report a Deny — a fail-open auth bug. The suggested alternative (memoize with an
  unset state) needs to mutate the receiver, which reintroduces exactly the race being removed elsewhere in this same
  change, since Policy values are shared read-only. The scan stays as the authority; the init() closes the gap for every
  policy this package owns.

klauspost:
@coderabbitai

  The statementClass recompute is a real staleness hazard — I confirmed it against the code: classify() returns the
  cached class when non-zero, and isAllowedFor derives ignoreResourceMatch from it, so replacing Actions in place on a
  statement inside an already-parsed policy can skip resource matching entirely. Fail-open. The finding is not wrong
  about the mechanism.

  I did not apply the fix, because I measured what it costs. Making classify() always call computeClass():

  ┌──────────────────────┬────────┬──────────────────┐
  │      benchmark       │ cached │ always recompute │
  ├──────────────────────┼────────┼──────────────────┤
  │ SingleStatementAllow │ 128 ns │ 206 ns (+61%)    │
  ├──────────────────────┼────────┼──────────────────┤
  │ DenyRule             │ 245 ns │ 404 ns (+65%)    │
  ├──────────────────────┼────────┼──────────────────┤
  │ WildcardMatching     │ 128 ns │ 333 ns (+160%)   │
  └──────────────────────┴────────┴──────────────────┘

  That deletes the reason this cache exists on the request path. And there's no cheap sound substitute:

  - Length-guarded invalidation misses same-size replacement — precisely the GetObjectAction swap the finding names — so
  it would read as a fix while leaving the hole open.
  - Classifying args.Action instead of the statement's set breaks two documented behaviors the finding asks me to
  preserve. It would kill the s3tables→S3 data path, which depends on the statement being classTable while the action is
  deliberately not a table action (class.has(classTable) && !TableAction(args.Action).IsValid()). And for legacy
  mixed-namespace statements — which statementClass's own comment says must "classify exactly as the individual
  predicates did" — it would start enforcing Resources where they were previously skipped, turning existing Allows into
  denies.

  So instead I documented the invariant where it's checkable, extending the class field comment: replacing Actions on a
  statement held by a parsed policy leaves it stale, a stale namespace can skip the resource check, and the remedy is
  Clone() or rebuilding the policy. The prior comment's "slower, never wrong" only covered the unpopulated case and was
  silent on mutation, which is what made the hazard invisible.

  Reaching this properly needs unexporting Actions or freezing parsed statements — a breaking API change, well outside
  this diff. If you want that hole actually closed rather than documented, say so and I'll scope it; the 60–160%
  regression is a decision for you, not one I should make silently.

klauspost:
@coderabbitai review

klauspost:
@coderabbitai The Reindex method was added for this. That is as far as we will go with `updateActionIndex `.

klauspost:
@coderabbitai review

policy/policy.go
```diff
@@ -510,12 +560,24 @@ func (iamp Policy) ValidateStrict() error {
...
+// derived state is what decides whether a statement's Resources are matched at
+// all, so leaving it stale can skip that check. It also moves a policy
+// assembled by hand onto the indexed evaluation path.
+func (iamp *Policy) Reindex() {
+	iamp.updateActionIndex()
+}
+
+// updateActionIndex fills the Action -> []Statements reverse lookup by
+// appending, so it discards the derived state first: a repeat call has to
+// replace the previous result rather than accumulate onto it.
```
taran-p: The "discards derived state first / repeat calls must replace not accumulate" reasoning reads better sitting on the two reset lines inside the function itself.

policy/admin-action.go
```diff
@@ -465,7 +474,25 @@ func (action AdminAction) Match(a AdminAction) bool {
 }
 
 // IsValid - checks if action is valid or not.
+//
+// The receiver is the pattern, so this asks whether the pattern matches any
+// supported admin action. Statement.isAdmin calls it for every action of every
+// statement it evaluates, so the two cases that can be answered without
+// touching SupportedAdminActions are answered first: a literal action can only
+// match by being in the set, and a pattern is only worth scanning when its
+// literal head is prefix-compatible with the admin namespace.
```
taran-p: Keep the first sentence; move the fast-path reasoning to the if blocks. Also, the Statement.isAdmin reference is stale — computeClass is the hot-path caller now.

policy/policy.go
```diff
@@ -117,17 +117,39 @@ func (a Args) GetRoleArn() string {
 }
 
 // Policy - iam bucket iamp.
+//
+// The fields below Statements are derived from it by updateActionIndex, so
+// changing Statements on a parsed policy leaves them stale. Call Reindex after
+// doing so.
```
taran-p: Move Reindex to the beginning of the comment and reasoning after. References to internal fields can be done in comments on those fields.

policy/policy.go
```diff
@@ -117,17 +117,39 @@ func (a Args) GetRoleArn() string {
...
+	// Whether any statement carries a Deny. See HasDenyStatement for why this
+	// is only a shortcut.
+	hasDeny bool
 }
 
-// HasDenyStatement returns if the policy has a deny statement.
+// HasDenyStatement returns if the policy has a deny statement. hasDeny is only
+// a shortcut, filled in by updateActionIndex; the statements stay authoritative
+// because a policy assembled directly never reaches that path, and trusting the
+// field alone would silently under-report its Deny.
```
taran-p: This and the hasDeny field comment point at each other. Keep the rationale on the field, leave this as the one-line behavior statement.

policy/statement.go
```diff
@@ -36,6 +36,18 @@ type Statement struct {
 	Resources    ResourceSet         `json:"Resource,omitempty"`
 	NotResources ResourceSet         `json:"NotResource,omitempty"`
 	Conditions   condition.Functions `json:"Condition,omitempty"`
+
+	// Namespace classification, filled in by Policy.updateActionIndex. Zero
+	// means "not classified yet" — classify() then derives it on the spot, so a
+	// statement built by a path that does not populate it is slower, never
+	// wrong. Statement.Clone deliberately leaves it zero.
```
taran-p: Trim to the invariant. What classify() does on zero belongs at classify(); the Clone note at Clone.

policy/statement.go
```diff
@@ -36,6 +36,18 @@ type Statement struct {
...
+	//
+	// Replacing Actions on a statement already held by a parsed policy leaves
+	// this stale, and the namespace decides whether Resources are matched at
+	// all, so a stale one can skip that check. Call Policy.Reindex after such a
+	// change; recomputing per evaluation instead costs 60-160% on the
+	// authorization path, which is why the value is cached at all.
```
taran-p: Benchmark numbers go in the PR description per AGENTS.md. "Cached because recomputing per evaluation is expensive" is enough.

policy/statement.go
```diff
@@ -177,6 +199,61 @@ func (statement Statement) validateActionTypes() error {
...
+func (statement Statement) classify() statementClass {
+	if statement.class != 0 {
+		return statement.class
+	}
+	return statement.computeClass()
+}
+
+// computeClass walks the action set once. The predicates it stands in for on
+// the authorization path each walked it separately, so a statement paid four
+// map iterator setups per evaluation to learn what one pass can tell it.
```
taran-p: First sentence is enough — the rest describes code this branch deletes.

wildcard/match.go
```diff
@@ -49,34 +79,55 @@ func Match(pattern, name string) (matched bool) {
...
-	for len(pattern) > 0 {
-		switch pattern[0] {
-		default:
-			if len(str) == 0 || str[0] != pattern[0] {
-				return false
+// deepMatchRune walks pattern against str one byte at a time, remembering only
+// the most recent '*' to resume from. Trying both alternatives at every '*'
+// instead — as a recursive matcher does — costs time exponential in the number
+// of stars, which a caller-supplied pattern can trigger: a policy action of
+// "*********x" kept AdminAction.IsValid busy for a minute.
```
taran-p:
```suggestion
// deepMatchRune matches pattern against str in a single pass, backtracking to
// the most recent '*'. Recursing on both branches at each '*' is exponential in
// the star count, which a caller-supplied pattern can trigger.
```
Tighten to two sentences and cut the "*********x" anecdote; the regression test pins that.

taran-p (approved):
Functionality looks good, tested with a few scenarios and they all were correct with improved performance.

I ran this through a personal Claude skill I have that checks for "Claudisms" in doc comments (describing internal workings rather than external behavior, changelog-style doc comments, etc.). All requested changes are nit style preference to change at your discretion and non-blocking.

rraulinio (approved):

policy/policy.go
```diff
@@ -117,17 +117,37 @@ func (a Args) GetRoleArn() string {
 }
 
 // Policy - iam bucket iamp.
+//
+// Reindex must be called after changing Statements on a policy that has already
+// been parsed, because the state derived from them is otherwise left stale.
```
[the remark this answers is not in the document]
klauspost: Your context isn't checking far enough back.

policy/statement.go
```diff
@@ -177,6 +194,60 @@ func (statement Statement) validateActionTypes() error {
...
+// time when there is one and deriving it on the spot otherwise — slower, never
+// wrong.
+func (statement Statement) classify() statementClass {
+	if statement.class != 0 {
+		return statement.class
+	}
+	return statement.computeClass()
+}
+
+// computeClass walks the action set once.
```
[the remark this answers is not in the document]
klauspost: I think it serves as good, to the point documentation.

PR #263, opened 2026-09-02
feat(dedup): add a bounded de-duplicating channel

harshavardhana:
## Motivation

Go channels cannot de-duplicate: there is no hook on send and no way to inspect a buffer. The single-key case is already idiomatic (`chan struct{}` of cap 1 plus a non-blocking send); this is the multi-key generalization, for coalescing repeated notifications about the same entity into a single unit of work.

`sync/dedup.Chan[K, T]` is a bounded channel holding at most one queued value per key. A key is reserved from enqueue until the value is received, so a duplicate is dropped only while its predecessor is still waiting to be consumed.

```go
c, err := dedup.NewChan(1024, func(e Event) string { return e.Bucket })
sent, err := c.Send(ctx, Event{Bucket: "photos"}) // sent == false, err == nil -> already queued
v, ok := c.Recv(ctx)                              // frees the key
```

## Design

- Three pieces of state keep the invariant *a key is pending iff its item is in the buffer*: a `slots` channel pre-filled with `size` tokens, the `items` buffer, and the pending-key map. A sender acquires a slot **before** touching the map, then double-checks the map under the mutex. That ordering matters: reserving the key first and then blocking on a full buffer lets a cancelled send drop a duplicate and enqueue nothing, silently losing an entry. Since a sender always holds a token, the buffer send and the slot return can never block.
- `items` is never closed. `Close` closes a separate channel (idempotent via `sync.Once`), so pending sends unblock with `ErrClosed` and `Recv` drains what is left before reporting false.
- API: `Send`/`TrySend` (`false, nil` = duplicate, plus `ErrFull` and `ErrClosed`), `Recv`/`TryRecv`, `Len`, `Close`. Entries carry their key through the buffer so `release` deletes the key that was inserted rather than recomputing it from a possibly mutated value.

## Validation

- `make test` (lint + `go test -race -tags kqueue ./...`) green.
- `go test -race -count=20 ./sync/dedup/` green.
- Tests cover: dedup while queued and re-send after receive, distinct keys queuing independently, backpressure, `TrySend` full, context cancellation on both sides (including that a cancelled send reserves nothing), close/drain/double-close, 32 goroutines racing on one key (exactly one wins, no slot leaked), and an 8-producer/4-consumer hammer asserting nothing is lost and all keys and slots are returned.

## Performance impact

New package, no existing code paths touched. Per send: one channel receive, one mutex-guarded map lookup, one buffered send.

## Documentation

No new environment variables or configuration flags, so `README.md` is unchanged; the package carries GoDoc.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a bounded channel that automatically drops duplicate queued values based on caller-defined keys.
  * Added blocking and non-blocking send and receive operations.
  * Added queue length reporting and safe, repeatable channel closure.
  * Added clear errors for closed and full channels.
  * Added context cancellation support for waiting operations.

* **Tests**
  * Added comprehensive coverage for deduplication, capacity limits, cancellation, closure, concurrency, and stress scenarios.


harshavardhana:
## Review feedback addressed

Applied 1 fix across 2 files from 1 review finding, and rebased onto `main`.

**Enqueue after `Close`** — valid. A sender holding a buffer slot could enqueue after `Close`, by which point a receiver may have observed closure over an empty buffer and stopped, leaving the value unreachable. `Close` now takes the same mutex as the enqueue path, so an enqueue either lands before closure is observable or returns its slot and reports `ErrClosed`. Added an interleaving test that races four senders against `Close` over 200 runs and asserts received count equals successful-send count, nothing is queued after close, and every slot comes back.

**Files changed:**
- `sync/dedup/dedup.go`
- `sync/dedup/dedup_test.go`

**Commit:** `7a36b03`

Also on the branch: `golang.org/x/crypto` bumped to v0.55.0, which clears the pre-existing govulncheck failure (GO-2026-6303, reached via `sftp`) that also fails on `main`.

**Validation:** `make test` (lint + `go test -race -tags kqueue ./...`) green; `go test -race -count=3 ./sync/dedup/` green.

**Deferred:** none.

harshavardhana:
@coderabbitai approve please

PR #267, opened 2026-09-18
fix(policy): resolve query-only condition keys from the query alone

harshavardhana:
## Summary

A condition key documented as coming from a query parameter **only** could be
satisfied by a request header instead. Those keys now resolve from the exact
name or not at all.

## What the fallback is for, and where it goes wrong

`getValuesByKey` looks a key up by `Name()` — the lowercase form left after the
service prefix is stripped — and falls back to the canonical header spelling:

```go
name := key.Name()                              // "s3:prefix" -> "prefix"
if values, found := m[name]; found { return values }
return m[http.CanonicalHeaderKey(name)]         // "Prefix"
```

That fallback is load-bearing for **header-sourced** keys. `s3:x-amz-acl`
becomes `x-amz-acl`, while a server records the header it came from
canonically as `X-Amz-Acl`, so without the fallback those keys never resolve.

Applied to a query-only key it is a bypass. Five keys document their value as
coming from a query parameter only:

| Key | `Name()` | fallback reached |
| --- | --- | --- |
| `s3:prefix` | `prefix` | `Prefix` |
| `s3:delimiter` | `delimiter` | `Delimiter` |
| `s3:max-keys` | `max-keys` | `Max-Keys` |
| `memory:prefix` | `prefix` | `Prefix` |
| `memory:max-keys` | `max-keys` | `Max-Keys` |

A server merges request headers into the condition values, so a caller can
supply `Prefix: allowed/` as a header with **no** `prefix` query parameter at
all. The condition is then evaluated against `allowed/` while the listing runs
with no prefix — the condition checking one value and the request being served
with another. Measured against a real `ListObjectsV2` request and a policy
whose `s3:ListBucket` grant is bounded by `StringLike s3:prefix`:

```
                                 | handler lists     | ListBucket allowed
query prefix inside the grant    | "allowed/"        | true    (correct)
query prefix outside the grant   | "other/"          | false   (correct)
HEADER prefix, no query prefix   | ""  (whole bucket)| true  <- before
HEADER prefix, no query prefix   | ""  (whole bucket)| false <- after
```

A present-but-empty `?prefix=` was already safe: the exact name is then found
and the fallback never runs. Only an entirely absent query parameter reached it.

## Changes

- `KeyName.IsQueryOnly()` plus the `queryOnlyKeys` set naming the five keys.
- `getValuesByKey` skips the canonical-header fallback for a query-only key.
  Header-sourced keys are untouched.

## Testing

`go build ./...` and `go test ./...` pass.

- `TestQueryOnlyKeyIgnoresAHeaderValue` covers all five keys against a header
  value, each query-only key against its exact query name, and two
  header-sourced keys in both spellings, so the fallback is pinned as still
  working where it is needed.
- `TestQueryOnlyKeysAreTheDocumentedSet` asserts the set both ways, so a key
  added to one and not the other fails.

Verified end to end against an unmodified AIStor build with a local `replace`:
the header-only row flips to denied and the query rows are unchanged.

## Scope

This changes condition evaluation for every policy using these keys, not one
feature, which is why it is separate from #266 — it is reviewable and
backportable on its own. Say the word and I will fold it into #266 instead.

## Summary by CodeRabbit

- **Bug Fixes**
  - Query-only condition values are now resolved exclusively from their exact query parameter names.
  - Prevented query-only conditions from incorrectly matching equivalent HTTP header names.
  - Preserved canonical header matching for conditions that support header-based values.
  - Added coverage to verify documented query-only keys and their lookup behavior.


harshavardhana:
Closing: not worth the blast radius.

The behaviour is real -- a `Prefix:` header satisfies a condition on
`s3:prefix` when the query parameter is absent -- but what it exposes is key
names, not object contents, and that exposure is already documented and
accepted in AIStor's agent design: the two resource-bound `Deny` statements
are what confine an agent on the object path, never the `s3:prefix`
condition.

Narrowing `getValuesByKey` changes condition evaluation for every policy in
the product that uses these five keys, which is a lot of surface to move for a
leak that is accepted by design where it matters.

PR #270, opened 2026-09-20
Fix policy de-duplication dropping statements on condition values

123456342g-lang:
## Motivation

`dropDuplicateStatements` discards statements that compare equal, and a `Deny` that goes with them is a restriction the caller wrote and no longer has. Two conditions that select different requests could render to the same string and be treated as duplicates.

`StringFunc.String` built its output as

```go
return fmt.Sprintf("%v:%v:%v", f.n, f.k, valueStrings)
```

and `%v` renders a `[]string` by joining it with spaces inside brackets, so the single value `"a b"` and the two values `{"a","b"}` both came out as:

```
StringEquals:s3:prefix:[a b]
```

A `Deny` restricted to `{"StringEquals":{"s3:prefix":"a b"}}` was therefore dropped as a duplicate of a `Deny` listing `"a"` and `"b"` as two values.

This is not confined to the hashed path as the `NotResources` omission in #261 was: `Functions.Equals`, which `Statement.Equals` calls, compares conditions through their `String` output as well, so policies of 10 statements or fewer — the exact-comparison path — collapsed the same pairs.

`IPAddrFunc.String` rendered its list the same way. Its values are CIDRs and cannot hold a separator, so no reachable collision is known there; it is changed to keep the two sibling renderings consistent.

## Changes

- `conditionString` renders a condition as a length-prefixed list of its name, key and values (`len:bytes,` per element), so equal output implies an equal list. `fmt` no longer decides whether two conditions differ.
- `Functions.String` and the two list-carrying functions go through it. The scalar functions (`Bool`, `Numeric`, `Date`, `Null`) print one value at a time and are untouched.
- Two tests: the renderings are injective over values holding a space, a comma, brackets and emptiness, and the de-duplication plus the resulting decision keep a `Deny` that differs only in its condition.

## Validation

- `go test -race ./...` — green.
- `golangci-lint run ./policy/...` — 0 issues.
- Reverting `StringFunc.String` to the previous form fails `TestStringFuncStringIsInjective` on its precondition (the two conditions compare equal again) and `TestDropDuplicateStatementsKeepsDistinctConditions` on the dropped statement, so the tests fail closed on the bug they describe.

## Compatibility

A policy that was relying on the collapse regains a statement. Where the regained statement is a `Deny`, a request that previously succeeded may start returning 403. That is the intended direction, but it belongs in release notes, as it did for the `NotResources` fix.

No API change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented distinct policy conditions from being treated as identical during comparison and statement deduplication.
  * Preserved separate policy statements when condition values differ in meaning but previously produced the same representation.
  * Ensured restrictive Deny decisions continue to apply correctly after duplicate-statement processing.

* **Tests**
  * Added coverage for distinct string and IP address condition representations.
  * Added regression tests for policy deduplication and decision evaluation.


harshavardhana (commented):
Reviewed this closely and reproduced the bug against the base commit — the diagnosis is right and the fix is correct. `encodeStringList`'s `len:value,` framing is genuinely injective, and folding the name and key in as length-prefixed elements makes it injective across function types too.

Confirming the collision is real, rendering at `525a565`:

```
["a","b","c"]  -> "StringEquals:s3:prefix:[a b c]"
"a b"          -> "StringEquals:s3:prefix:[a b]"
["a","b"]      -> "StringEquals:s3:prefix:[a b]"   <- same string, opposite semantics
```

Four things inline before merge. The substantive one is `TestDistinctConditionsKeepRestrictingTheDecision`, which currently passes on unfixed code — CodeRabbit already flagged it and is right; I've added the reproduction to that thread.

Off-diff notes, nothing for you to change here:

- `dropDuplicateStatementsMany` and `MergePolicies` both intend a random hash seed and never get one. `rand.Read(baseSeed[:])` fills 8 random bytes, then `binary.LittleEndian.PutUint64(baseSeed[:], seed)` writes the still-zero `seed` back over them — the argument is inverted, it wanted `seed = binary.LittleEndian.Uint64(baseSeed[:])`. Both sites, `policy/policy.go:448` and `:474`. The seed is therefore always zero and the dedup hash is fully deterministic, which is exactly what the randomization existed to prevent. Predates this PR; I'm fixing it separately. Mentioning it because it is the same comparator you are working on.
- `MergePolicies` is the more reachable path for what you found: it hash-dedupes statements across separately attached policies, so the two colliding statements only need to come from two different policies rather than from one hand-written one. Worth covering if you want one more test.
- `Functions.Equals` has a pre-existing multiset bug — `[X,X]` compares equal to `[X,Y]`. Unreachable from JSON parsing (an object cannot hold two identical name/key pairs), reachable via `NewFunctions`. Out of scope, noting it while we are here.
- `Statement.Equals` is exported and becomes stricter with this change. Callers using it as a "policy unchanged" check may now see churn on policies that previously compared equal — worth a line in the Compatibility section.

policy/condition/func.go
```diff
@@ -128,15 +129,41 @@ func (functions Functions) MarshalJSON() ([]byte, error) {
...
 	for _, f := range functions {
-		s := fmt.Sprintf("%v", f)
-		funcStrings = append(funcStrings, s)
+		funcStrings = append(funcStrings, f.String())
 	}
 	sort.Strings(funcStrings)
 
-	return fmt.Sprintf("%v", funcStrings)
+	return encodeStringList(funcStrings)
 }
```
harshavardhana, line 167: This aggregate `Functions.String()` is not load-bearing for the fix, and I would revert it.

Neither comparator uses it. `Functions.Equals` compares the *per-function* `String()` (`func.go:100`), and `Statement.hash` hashes the *per-condition* `String()` (`statement.go:720`). Grepping this repo, the only other consumer is a `%v` in a test error message at `policy/policy_test.go:2209`.

What the change does do is turn an exported Stringer into a wire format: `[StringEquals:s3:prefix:[a]]` becomes `28:StringEquals:s3:prefix:[a],`. That is exported surface, so minio/minio or console may well be formatting it into logs or diagnostics.

The `Function` interface is already sealed — `evaluate`, `key`, `name`, `toMap` and `clone` are all unexported — so adding an unexported `canonicalKey() string` and pointing `Functions.Equals` and `Statement.hash` at *that* costs nothing in compatibility, keeps `String()` human-readable, and gives the injectivity invariant a name.

policy/condition/func.go
```diff
@@ -128,15 +129,41 @@ func (functions Functions) MarshalJSON() ([]byte, error) {
...
+		b = append(b, ',')
+	}
+	return string(b)
+}
+
+// conditionString renders a condition as a string that is injective in its
+// name, its key and its values. Functions.Equals and Statement.hash compare
+// conditions through this string, and dropDuplicateStatements discards whatever
+// compares equal, so a rendering that merges two conditions silently throws
+// away a statement the caller wrote -- a Deny among them.
```
harshavardhana, line 151: Two things here.

Comment length: repo convention (CLAUDE.md) is minimal comments explaining *why*, not *what*. This block and the two at the top of the new test files read as narrative history. Something like "Injective in name, key and values: `Functions.Equals` and `Statement.hash` compare through this, and `dropDuplicateStatements` drops whatever compares equal." carries the same load in two lines.

Coverage of the invariant: only `stringFunc` and `ipaddrFunc` move to this encoding. `booleanFunc`, `nullFunc`, `numericFunc` and `dateFunc` still use `fmt.Sprintf`. Not a live collision today — those are single-valued, and old-format strings start with a letter while new-format ones start with a digit, so the two formats cannot alias — but the invariant is not stated anywhere, so the next person adding a multi-value condition will copy the `fmt.Sprintf` pattern and quietly reintroduce this. One line on the interface method (or on the proposed `canonicalKey()`) would pin it.

policy/condition/func_string_injective_test.go
```diff
@@ -0,0 +1,153 @@
+package condition
```
harshavardhana, line 1: Missing the AGPL copyright header that every other `.go` file in the repo carries. `golangci-lint` has no `goheader` linter enabled here, so CI will not flag it.

policy/condition/func_string_injective_test.go
```diff
@@ -0,0 +1,153 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// Sibling functions carry the same rendering contract, so pin ipaddrFunc too:
+// its values are CIDRs and cannot hold a separator, but it must stay injective
+// regardless.
+func TestIPAddrFuncStringIsInjective(t *testing.T) {
```
harshavardhana, line 126: This one passes on unfixed code. Running it at `525a565`:

```
--- FAIL: TestStringFuncStringIsInjective
--- FAIL: TestStringFuncStringKeepsDistinctValuesDistinct
--- PASS: TestIPAddrFuncStringIsInjective
```

The two `stringFunc` tests correctly fail, this one is green — the three CIDRs cannot collide under the old `%v` rendering either, as the PR description itself notes. Either drop it or give it a case that actually discriminates.

policy/condition_dedup_regression_test.go
```diff
@@ -0,0 +1,129 @@
+package policy
```
harshavardhana, line 1: Same here — missing the AGPL copyright header.

policy/condition_dedup_regression_test.go
```diff
@@ -0,0 +1,129 @@
...
+// The same collapse reaches the decision, not just the statement list: the
+// dropped Deny stopped restricting a request that the parser had been told to
+// restrict.
+func TestDistinctConditionsKeepRestrictingTheDecision(t *testing.T) {
+	const src = `{
+	  "Version": "2012-10-17",
+	  "Statement": [
+	    {"Effect": "Allow", "Action": ["s3:GetObject"], "Resource": ["arn:aws:s3:::bucket/*"]},
+	    {"Effect": "Deny", "Action": ["s3:GetObject"], "Resource": ["arn:aws:s3:::bucket/*"],
+	     "Condition": {"StringEquals": {"s3:prefix": "a b"}}},
```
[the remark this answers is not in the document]
harshavardhana: Concur — this is the one to act on.

As it stands the fixture has a single conditional Deny and nine unconditional Allows on distinct resources, so nothing collides and `dropDuplicateStatements` has nothing to drop, before or after the fix. At `525a565` (unfixed) it is green:

```
--- PASS: TestDistinctConditionsKeepRestrictingTheDecision
```

With the pairing suggested here — a `{"StringEquals": {"s3:prefix": ["a", "b"]}}` Deny immediately before the `"a b"` one — it reproduces properly at `525a565`:

```
condition_dedup_regression_test.go:118: the Deny on prefix "a b" was dropped: the request it refuses was allowed
--- FAIL: TestDistinctConditionsKeepRestrictingTheDecision
```

and passes at `588a58d`.

Worth doing for a second reason: it takes the fixture to 11 statements, crossing the `> 10` threshold in `dropDuplicateStatements`, so it exercises `dropDuplicateStatementsMany` (the hashed path) rather than only `dropDuplicateStatementsOriginal`. Right now the decision-level test never reaches the hashed path at all.

